PR #8, opened 2023-04-11
[DEV-1462] store redirect uri

Krysh90:


src/screens/buy/payment.screen.tsx
```diff
@@ -127,7 +127,7 @@ export function BuyPaymentScreen(): JSX.Element {
   return (
     <Layout
       backTitle={showsCompletion ? translate('screens/buy/payment', 'Done!') : translate('screens/buy/payment', 'Buy')}
-      appPage={AppPage.BUY}
+      appPage={showsCompletion ? AppPage.BUY : undefined}
```
davidleomay, line 130: What is this exactly?

davidleomay (approved):

src/screens/buy/payment.screen.tsx
Krysh90: if undefined a back icon will be displayed. That something nobody noticed, but just saw it while testing. on payment screen (amount input) we weren't showing the back icon

PR #12, opened 2023-04-13
[DEV-1460] added profile screen, stub for sell screen and kyc data ha…

Krysh90:
…ndling


src/screens/profile.screen.tsx
```diff
@@ -50,6 +52,13 @@ export function ProfileScreen(): JSX.Element {
 
     mail: [Validations.Required, Validations.Mail],
     phone: [Validations.Required, Validations.Phone],
+
+    organizationName: Validations.Required,
+    organizationStreet: Validations.Required,
+    organizationHouseNumber: Validations.Required,
+    organizationLocation: Validations.Required,
+    organizationZip: Validations.Required,
+    organizationCountry: Validations.Required,
```
davidleomay, line 61: These should only be required, if accountType !== Personal

davidleomay (approved):

src/screens/profile.screen.tsx
Krysh90: thought so too, but the fun thing. if there is no input for this, it can't be required. I have looked into payment and there it is the same. I have tested it, it does work

davidleomay: Ah true, because the input fields are not rendered at all, you're right 🤔

PR #15, opened 2023-04-18
[DEV-1480] min deposit changes

Krysh90:


src/screens/buy/payment.screen.tsx
```diff
@@ -104,7 +104,12 @@ export function BuyPaymentScreen(): JSX.Element {
       purpose: buy.remittanceInfo,
       isSepaInstant: buy.sepaInstant,
       recipient: `${buy.name}, ${buy.street} ${buy.number}, ${buy.zip} ${buy.city}, ${buy.country}`,
+      estimatedAmount: translate('screens/buy/payment', '≈ {{estimatedAmount}} {{asset}}', {
+        estimatedAmount: '' + buy.estimatedAmount,
+        asset: asset?.name ?? '',
+      }),
```
davidleomay, line 110: I think the translation should be there, where it the estimatedAmount is used (similar to minFee).

davidleomay (approved):

PR #16, opened 2023-04-19
[DEV-1480] added amount spinner and added buyable filter

Krysh90:


src/api/contexts/asset.context.tsx
```diff
@@ -40,9 +40,14 @@ export function AssetContextProvider(props: PropsWithChildren): JSX.Element {
...
 
-  function getAsset(id: number): Asset | undefined {
+  function getAsset(id: number, filter?: { buyable?: boolean; sellable?: boolean }): Asset | undefined {
     return Array.from(assets.values())
       .reduce((prev, curr) => prev.concat(curr), [])
+      .filter((asset) => {
+        if (!filter) return true;
+        if (filter.buyable) return filter.buyable === asset.buyable;
+        if (filter.sellable) return filter.sellable === asset.sellable;
+      })
```
davidleomay, line 50: If `filter.buyable` is false, this will return an empty list. If `filter.buyable` and `filter.sellable` are true, it will still return only all buyables.
If you only want to be able to filter by true, you can do it like: `.filter((asset) => (!filter.buyable || asset.buyable) && (!filter.sellable || asset.sellable))` with `&&` or `||` depending on what you want 😁

davidleomay (approved):

PR #18, opened 2023-04-28
[DEV-1493] only enabling sell button if balance is available

Krysh90:


src/contexts/balance.context.tsx
```diff
@@ -43,6 +44,7 @@ export function BalanceContextProvider(props: PropsWithChildren): JSX.Element {
     () => ({
       balances,
       readBalances,
+      hasBalance: balances?.find((balance) => +balance.amount > 0) !== undefined,
```
davidleomay, line 47:
```ts
balances?.some((balance) => +balance.amount > 0)
```

davidleomay (approved):

PR #94, opened 2023-09-13
[DEV-1755] Deactivate sell with wallet connect

Kolibri1990:


src/config/feature-tree.ts
davidleomay, line 1: `walletconnect` needs to be removed for `shapeshift-sell-ethereum` (and below)

davidleomay (approved):

PR #115, opened 2023-10-31
[DEV-1679], [DEV-1880], [DEV-1692], [DEV-1755], [DEV-1893], [DEV-1872], [DEV-1790], [DEV-1909], [DEV-1782], [DEV-1880], [DEV-1887], [DEV-1757] & [DEV-1826]

Kolibri1990:
- Added sell info for all chains
- Increased QR size
- Chunking
- Walled Connect embedded setup
- Fixed WC responsive
- Added max button
- Using new payment info error field
- Added fallback for Metamask
- Login Button & My DFX
- Improved Taro UI
- Type and Address selection (HW wallets)
- Discount code param
- Add QR code to payment request


.env.loc
```diff
@@ -1,4 +1,5 @@
 PORT=3001
+PUBLIC_URL=http://localhost:4000
```
bernd2022, line 2: Why "localhost:4000" and PORT=3001 - shouldn't it be "localhost:3001"?

bernd2022 (approved):

.env.loc
davidleomay: `PUBLIC_URL` is used for the widget "deployment", which is locally served under port 4000 (see `package.json` > `scripts.serve`)

PR #119, opened 2023-11-02
[DEV-1922] Address selection trezor

Kolibri1990:


src/hooks/wallets/trezor.hook.ts
```diff
@@ -63,15 +76,44 @@ export function useTrezor(): TrezorInterface {
...
+            ? await TrezorConnect.getAddress({
+                path: KeyPath.BTC(bitcoinAddressType).address(i),
+                showOnTrezor: false,
+              })
+            : await TrezorConnect.ethereumGetAddress({ path: KeyPath.ETH.address(i), showOnTrezor: false });
+        if (result.success) addresses.push(result.payload.address);
+      }
+      return addresses;
+    } catch (e) {
+      throw e;
```
bernd2022, line 99: Does it make sense to catch an exception and then throw it right back?

Shouldn't a message come to the user here so that he sees that a problem has occurred?

bernd2022 (approved):

src/hooks/wallets/trezor.hook.ts
davidleomay: No, doesn't make sense, the try-catch can be removed.

bernd2022: The try-catch is also in the other hardware wallet hooks and can therefore be removed there as well.

src/hooks/wallets/trezor.hook.ts
```diff
@@ -63,15 +76,44 @@ export function useTrezor(): TrezorInterface {
...
+      for (let i = startIndex; i < startIndex + count; i++) {
+        const result =
+          wallet === WalletType.TREZOR_BTC
+            ? await TrezorConnect.getAddress({
+                path: KeyPath.BTC(bitcoinAddressType).address(i),
+                showOnTrezor: false,
+              })
+            : await TrezorConnect.ethereumGetAddress({ path: KeyPath.ETH.address(i), showOnTrezor: false });
+        if (result.success) addresses.push(result.payload.address);
+      }
```
davidleomay, line 96: Maybe we should throw an error, if it wasn't a success?

src/hooks/wallets/trezor.hook.ts
davidleomay: For BitBox it's already removed in https://github.com/DFXswiss/services/pull/120

davidleomay (approved):

PR #131, opened 2023-11-09
lightning und co

TaprootFreak:


davidleomay (approved):
CLI does not support Lightning

PR #238, opened 2024-05-13
[DEV-2528] Show tx fee rate [DEV-2532] Adapted to API

davidleomay:


src/translations/languages/de.json
Yannick1712, line 1: I think we have to rename the Description field + value to message or?

src/translations/languages/fr.json
Yannick1712, line 1: Same here

src/translations/languages/it.json
Yannick1712, line 1: Same here

src/translations/languages/de.json
davidleomay: The translated string is still `Description`:
image
Do you think we should rename it to `Message`?

Yannick1712: Hm yeah we could leave it like that. Don't know first thought I think I would prefer message but we can also take the description 👍

PR #245, opened 2024-05-23
[DEV-2570] enable standalone services login module

lapatric:


src/contexts/app-handling.context.tsx
```diff
@@ -32,6 +32,7 @@ export interface AppParams {
   signature?: string;
   mail?: string;
   wallet?: string;
+  wallets?: string;
   refcode?: string;
```
davidleomay, line 36: You need to add this also to `extractUrlParams` and `index-widget.tsx`

src/screens/home.screen.tsx
```diff
@@ -112,6 +113,12 @@ export function HomeScreen(): JSX.Element {
         navigate('/');
 
       default:
+        if (appParams.wallets) {
+          console.log("Closing services");
```
davidleomay, line 117: I wouldn't use the wallets param as indicator if we should close the services. Can we somehow use the service parameter for this?

davidleomay (approved):

PR #264, opened 2024-06-13
keep menu button when headless

lapatric:


src/components/navigation.tsx
```diff
@@ -29,6 +29,7 @@ interface BackButtonProps extends PropsWithChildren {
 }
 
 interface NavigationIframeProps extends BackButtonProps {
+  headless?: boolean;
   isOpen: boolean;
   setIsOpen: (value: SetStateAction<boolean>) => void;
 }
```
davidleomay, line 35: I would just use the `appHandlingContext` in the Navigation component instead of passing it as prop

davidleomay (approved):

PR #276, opened 2024-06-19
[DEV-2568] Dubai transactions

lapatric:


src/screens/sell-info.screen.tsx
davidleomay, line 1: Add translations?

davidleomay (approved):

PR #279, opened 2024-06-20
[DEV-2616] Auto-refresh transaction status screen

lapatric:


src/screens/transaction.screen.tsx
```diff
@@ -67,10 +67,18 @@ export function TransactionStatus(): JSX.Element {
...
-      getTransactionByUid(id)
-        .then(setTransaction)
-        .catch((error: ApiError) => setError(error.message ?? 'Unknown error'));
+    const fetchTransaction = () => {
+      if (id && transaction?.state !== TransactionState.COMPLETED) {
+        getTransactionByUid(id)
+          .then(setTransaction)
+          .catch((error: ApiError) => setError(error.message ?? 'Unknown error'));
+      }
+    };
```
davidleomay, line 76: Do you see the state change of transaction in this callback?

lapatric: Good point, I've added `transaction?.state` to the dependency list.

davidleomay (approved):

PR #291, opened 2024-06-28
[DEV-26146 Support improvements

davidleomay:


TaprootFreak (approved):

src/screens/support-issue.screen.tsx
```diff
@@ -71,29 +76,33 @@ export function SupportIssueScreen(): JSX.Element {
     reasons.length === 1 && setValue('reason', reasons[0]);
   }, [reasons]);
 
+  useEffect(() => {
+    if (selectedType === SupportIssueType.TRANSACTION_ISSUE) {
+      if (!id) navigate('/support/issue/tx');
```
lapatric, line 81: To improve UX for the case that the user selects "Issue type" last, one could carry over any data the user already provided to the new `support/issue/tx/:id` screen. But not sure how easy this is ^^'. It seems to work the other way around already, i.e. from `support/issue/tx/:id` to `support/issue/`.

lapatric (approved):

src/screens/support-issue.screen.tsx
davidleomay: Possible, but we would need to store it in a context somewhere 🤔  it's probably currently not worth the effort, considering that the type selection is at the top 😅

PR #294, opened 2024-06-28
[NO-TASK] update nav menu links

lapatric:


src/screens/swap.screen.tsx
```diff
@@ -400,7 +400,7 @@ export default function SwapScreen(): JSX.Element {
   });
 
   return (
-    <Layout title={translate('general/services', 'Swap')} textStart rootRef={rootRef}>
+    <Layout title={translate('navigation/links', 'Swap')} textStart rootRef={rootRef}>
```
davidleomay, line 403: Hm, why this?

davidleomay (approved):

src/screens/swap.screen.tsx
lapatric: The difference is "Tauschen" vs "Swap" for example. `navigation/links` translates to "SWAP" for all languages and is consistent with the "Swap" label used in the nav menu. So I thought it made sense to use "Swap" for the title too

PR #305, opened 2024-07-03
[DEV-2716] allow dynamic address length via blankedAddress()

lapatric:


src/screens/transaction.screen.tsx
davidleomay, line 1: Use dynamic length

src/util/utils.ts
```diff
@@ -25,9 +25,16 @@ export function isNode(e: EventTarget | null): e is Node {
...
 }
 
-export function blankedAddress(address: string, displayLength = 24): string {
-  return address.length > displayLength
-    ? `${address.slice(0, displayLength / 2)}...${address.slice(address.length - displayLength / 2)}`
+export function blankedAddress(
+  address: string,
+  { displayLength = 24, dynamicLength = false }: { displayLength?: number; dynamicLength?: boolean } = {},
+): string {
+  if (dynamicLength) displayLength = Math.min(Math.floor((window.innerWidth * 0.5) / 10), address.length);
```
davidleomay, line 32: Use component width

src/App.tsx
```diff
@@ -170,9 +170,10 @@ export interface WidgetParams extends AppParams {
 interface AppProps {
   routerFactory: (routes: RouteObject[]) => Router;
   params?: WidgetParams;
+  width?: number;
 }
 
-function App({ routerFactory, params }: AppProps) {
+function App({ routerFactory, params, width }: AppProps) {
   const router = routerFactory(Routes);
```
davidleomay, line 177: Why not just measure the width with the resize observer hook?

PR #492, opened 2024-12-03
[DEV-3385] Open file directly under /file (#490)

lapatric:
* [DEV-3385] Open file directly under /file


src/util/utils.ts
davidleomay, line 1: Open in same tab, if auto-open

davidleomay (approved):

PR #518, opened 2025-01-09
[DEV-3506] new DataRequest support issue reason

lapatric:
- [DEV-3506] new DataRequest support issue reason

[DEV-3506]: https://dfxswiss.atlassian.net/browse/DEV-3506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
       "version": "1.0.3",
       "license": "MIT",
       "dependencies": {
-        "@dfx.swiss/react": "^1.3.0-beta.111",
-        "@dfx.swiss/react-components": "^1.3.0-beta.111",
+        "@dfx.swiss/react": "^1.3.0-beta.113",
+        "@dfx.swiss/react-components": "^1.3.0-beta.113",
         "@ledgerhq/hw-app-btc": "^6.24.1",
         "@ledgerhq/hw-app-eth": "^6.33.7",
         "@ledgerhq/hw-transport-webhid": "^6.27.19",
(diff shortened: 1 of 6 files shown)


davidleomay (commented):
Shouldn't we filter it out from issue type dropdown (see ticket)?

lapatric:
> Shouldn't we filter it out from issue type dropdown (see ticket)?

Yes, but I believe that is already taken care of through the [`IssueReasons`](https://github.com/DFXswiss/services/blob/a1d4a21ac5b30dd4d6de31521696088d136c4f8c/src/screens/support-issue.screen.tsx#L49) mapping in `support-issue.screen.tsx`.

lapatric:
> Shouldn't we filter it out from issue type dropdown (see ticket)?

@davidleomay You're right, but I believe that is already taken care of through the [`IssueReasons`](https://github.com/DFXswiss/services/blob/a1d4a21ac5b30dd4d6de31521696088d136c4f8c/src/screens/support-issue.screen.tsx#L49) mapping in `support-issue.screen.tsx`.

davidleomay:
> > Shouldn't we filter it out from issue type dropdown (see ticket)?
> 
> @davidleomay You're right, but I believe that is already taken care of through the [`IssueReasons`](https://github.com/DFXswiss/services/blob/a1d4a21ac5b30dd4d6de31521696088d136c4f8c/src/screens/support-issue.screen.tsx#L49) mapping in `support-issue.screen.tsx`.

True 😉

davidleomay (approved):

PR #572, opened 2025-05-07
[DEV-3884], [DEV-3899], [DEV-3915] add TX receipt button, display external ID, add chargebackAsset

lapatric:


src/util/utils.ts
```diff
@@ -8,9 +8,11 @@ export function isEmpty(val: any): boolean {
   return val === undefined || val === '' || val === null || (Array.isArray(val) && val.length === 0);
 }
 
-/**
- * Returns the new value if it's different from the default, otherwise undefined
- */
+export function removeNullFields<T extends Record<any, any>>(entity?: T): Partial<T> {
+  if (!entity) return {} as Partial<T>;
```
davidleomay, line 12: Should we `return entity;`, if it's not defined?

davidleomay (approved):
- [x] Wait for API PR

davidleomay (approved):

PR #579, opened 2025-05-12
[DEV-3847] set payment currency from route

lapatric:


lapatric:
- Await https://github.com/DFXswiss/api/pull/2127

davidleomay (approved):

PR #588, opened 2025-05-19
[DRAFT]: Add deeplink button

Danswar:


src/screens/payment-link.screen.tsx
```diff
@@ -547,6 +551,39 @@ export default function PaymentLinkScreen(): JSX.Element {
     }
   }
 
+  async function payWithBinancePay() {
+    if (!payRequest?.id) return;
+    try {
+      const { uri: deeplink } = await fetchJson(
+        `${process.env.REACT_APP_API_URL}/v1/lnurlp/${payRequest.id}?method=BinancePay&asset=USDT`,
+      );
```
davidleomay, line 559: Call the callback URL
`url(payRequest.callback, new URLSearchParams({ quote: payRequest.quote.id, method: 'BinancePay', asset: 'USDT'})`

src/screens/payment-link.screen.tsx
```diff
@@ -547,6 +551,39 @@ export default function PaymentLinkScreen(): JSX.Element {
...
+      window.location.href = deeplink;
+    } catch (_) {}
+  }
```
davidleomay, line 562: `setError`?

src/screens/payment-link.screen.tsx
```diff
@@ -547,6 +551,39 @@ export default function PaymentLinkScreen(): JSX.Element {
...
+      ? [
+          {
+            name: C2BPaymentProviders.BINANCE_PAY,
+            iconUrl: 'https://content.dfx.swiss/img/v1/services/wallets/BinanceApp.webp',
+            onClick: payWithBinancePay,
+          },
+        ]
+      : []),
+  ];
+
```
lapatric, line 586: Instead of defining the BinancePay wallet separately out here, I would merge it all into the `CompatibleWallets` object in `payment-link-wallets.ts`, so that we all have it in one place. We can rename `CompatibleWallets` to simply `PaymentLinkWallets` or something. And then define the mapping to an extended `WalletMeta` interface, like so:

```typescript
export interface WalletMeta {
  iconUrl: string;
  websiteUrl?: string;
  onClick?: (payRequest) => Promise<void>;
  recommended?: boolean;
  semiCompatible?: boolean;
}
```

We could also make the `PaymentLinkWallets` object an array of `WalletMeta` items instead and add the `name` field to the `WalletMeta` interface, maybe that's better.

lapatric: Also note, that `BtcTaro` and `'Cake Wallet'` in the `CompatibleWallets` object are missing the `recommended: true` field. That was like that before your changes already but it would be good if you could add it :)

lapatric: Then to obtain the recommended and semi-compatible wallets you can just use the `recommended` and `semiCompatible` fields of the new `WalletMeta` interface. You can then get rid of the `RecommendedWallets` array in `payment-link-wallets.ts` if you like, just make sure the order is preserved.

Danswar:
Thanks for the comments guys, we can close this draft in favor of https://github.com/DFXswiss/services/pull/596.

PR #601, opened 2025-05-29
[DEV-4017]: Muun wallet

Danswar:
- Added Muun the list, it works querying the PR directly as it is not 100% compatible with our LNURLs.
- Deeplinking logic is growing in complexity so I moved it out into a new hook.
- Redirection was too fast, the completion screen is almost not noticeable, added 2 secs delay.


Danswar:
@lapatric I can't assign you as a reviewer so I'm tagging you here :)

lapatric (approved):

PR #614, opened 2025-06-16
[NO-TASK]: Manage payment link with access key

Danswar:
https://github.com/user-attachments/assets/4d39e474-72d7-4be0-9643-254923e79f97

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,7 +30,7 @@ const KycLogScreen = lazy(() => import('./screens/kyc-log.screen'));
 const LinkScreen = lazy(() => import('./screens/link.screen'));
 const PaymentRoutesScreen = lazy(() => import('./screens/payment-routes.screen'));
 const PaymentLinkScreen = lazy(() => import('./screens/payment-link.screen'));
-const PaymentLinkManageScreen = lazy(() => import('./screens/payment-link-manage.screen'));
+const PaymentLinkPosScreen = lazy(() => import('./screens/payment-link-pos.screen'));
 const InvoiceScreen = lazy(() => import('./screens/invoice.screen'));
 const SellInfoScreen = lazy(() => import('./screens/sell-info.screen'));
 const SupportIssueScreen = lazy(() => import('./screens/support-issue.screen'));
@@ -132,8 +132,8 @@ export const Routes = [
         element: withSuspense(<PaymentLinkScreen />),
       },
       {
-        path: 'manage',
-        element: withSuspense(<PaymentLinkManageScreen />),
+        path: 'pos',
+        element: withSuspense(<PaymentLinkPosScreen />),
       },
     ],
   },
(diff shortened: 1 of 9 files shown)


src/contexts/payment-link.context.tsx
```diff
@@ -125,7 +135,7 @@ export function PaymentLinkProvider(props: PropsWithChildren): JSX.Element {
     if (lightningUrlParam) {
       apiUrl = Lnurl.decode(lightningUrlParam);
       setParams({ lightning: undefined });
-    } else if (urlParams.size) {
+    } else if (isPaymentInvoiceRequest(urlParams)) {
```
Danswar, line 138: @lapatric I added this new thing here, is it sufficient for its previous use case? or are there more things that depends on urlParams.size besides creating invoice payments?

src/App.tsx
```diff
@@ -120,11 +121,21 @@ export const Routes = [
...
     ),
+    children: [
+      {
+        index: true,
+        element: withSuspense(<PaymentLinkScreen />),
+      },
+      {
+        path: 'manage',
+        element: withSuspense(<PaymentLinkManageScreen />),
+      },
```
davidleomay, line 137: What would you think of `pos` as a path?

Danswar: thanks! that's a way better name

Danswar: done!

src/screens/payment-link-pos.screen.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+function CreatePaymentForm({ payRequest, paymentLinkApiUrl, fetchPayRequest }: PaymentFormProps): JSX.Element {
+  const { translate } = useSettingsContext();
+  const { route, key } = useAppParams();
+  const { call } = useApi();
+  const [isLoading, setIsLoading] = useState(false);
+
+  const rules = Utils.createRules({
+    paymentAmount: [Validations.Required, Validations.Custom((value) => !isNaN(Number(value)) || 'invalid amount')],
+  });
```
davidleomay, line 89: We could use `!isNaN(Number(value)) || 'pattern')`, this is already translated

src/screens/payment-link-pos.screen.tsx
```diff
@@ -0,0 +1,346 @@
...
+    control,
+    handleSubmit,
+    formState: { errors },
+  } = useForm<{ amount: number }>({
+    mode: 'onTouched',
+  });
+
+  return (
+    <Form control={control} rules={rules} onSubmit={handleSubmit(createPayment)} errors={errors}>
+      <StyledVerticalStack gap={6} full>
```
davidleomay, line 127: Add `translate={translateError}` to translate the errors (with `const { translateError } = useSettingsContext();`)

src/screens/payment-link-pos.screen.tsx
```diff
@@ -0,0 +1,346 @@
...
+          fontWeight: 600,
+          fontSize: size * 0.3,
+          userSelect: 'none',
+        }}
+      >
+        {timeLeft}
+      </span>
+    </div>
+  );
+}
```
davidleomay, line 346: Move this to separate file in `/components`?

lapatric (approved):

PR #677, opened 2025-08-14
Fix: Login button visibility issue in navigation menu

TaprootFreak:
## Problem Description

The login button in the navigation menu was **completely hidden** for users when accessing the application through certain screens (payment-link, payment-link-pos, payment-link-assign), making it **impossible for users to log in**.

### Root Cause Analysis

1. **Hidden Condition Issue**: The StyledButton component had a `hidden={small && !isLoggedIn}` prop that would hide the button when:
   - The menu was in "small" mode (`small=true`) 
   - AND the user was not logged in (`!isLoggedIn`)

2. **Small Menu Mode**: The following screens set `smallMenu: true` in their layout configuration:
   - `payment-link.screen.tsx` (line 235)
   - `payment-link-pos.screen.tsx` (line 45)  
   - `payment-link-assign.screen.tsx` (line 20)

3. **User Impact**: When anonymous users accessed these payment-related screens, they had **no way to log in** because the login button was completely hidden from the UI.

### Navigation Issue

Additionally, the navigation to `/login` was not working properly due to URL parameter preservation:
- The app was likely embedded or accessed with query parameters
- The React Router's `navigate()` function preserves these parameters by default
- This caused the navigation to `/login` to fail silently

### 1. Remove Hidden Condition
```diff
- hidden={small && !isLoggedIn}
```
This ensures the login/logout button is **always visible** regardless of menu size or authentication state.

### 2. Fix Navigation
```javascript
window.location.href = window.location.origin + '/login';
```
Using `window.location` ensures:
- Complete page navigation without React Router
- All URL parameters are cleared
- Clean navigation to the login page

## Testing

Tested on:
- Payment link screens with small menu enabled
- Regular screens with full menu
- Both logged-in and logged-out states
- With and without URL parameters

## Impact

- **Critical Fix**: Restores ability for users to log in from payment screens
- **Improved UX**: Users always have access to authentication controls
- **No Breaking Changes**: Existing functionality remains intact

This is a critical bug fix that affects user accessibility to core authentication features.

### DIFF
--- a/src/components/navigation.tsx
+++ b/src/components/navigation.tsx
@@ -109,13 +109,12 @@ function MenuIcon({ icon, setIsNavigationOpen }: IconContentProps): JSX.Element 
 }
 
 function NavigationMenu({ setIsNavigationOpen, small = false }: NavigationMenuContentProps): JSX.Element {
-  const { navigate } = useNavigation();
   const { translate } = useSettingsContext();
   const { user } = useUserContext();
   const { isLoggedIn, logout: apiLogout } = useSessionContext();
 
   async function login() {
-    navigate('/login');
+    window.location.href = window.location.origin + '/login';
     setIsNavigationOpen(false);
   }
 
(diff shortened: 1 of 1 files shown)


src/components/navigation.tsx
```diff
@@ -238,7 +237,6 @@ function NavigationMenu({ setIsNavigationOpen, small = false }: NavigationMenuCo
             onClick={isLoggedIn ? logout : login}
             color={StyledButtonColor.STURDY_WHITE}
             width={StyledButtonWidth.FULL}
-            hidden={small && !isLoggedIn}
           />
```
davidleomay, line 240: Revert this

PR #699, opened 2025-08-28
Separate blockchain and address selection into individual dropdowns

NoahPay:
- Split combined dropdown into two separate components (BlockchainSelector and AddressSelector)
- Position dropdowns side-by-side with consistent sizing matching existing form fields
- Set current address as default selection to improve user experience
- Prevent payment information display until address is selected
- Maintain consistent layout across buy and order interfaces


TaprootFreak:
Sieht dann so aus:

Bildschirmfoto 2025-08-28 um 14 30 06

PR #772, opened 2025-12-09
Add SafeAccount multi-account system

TaprootFreak:
## Summary
- Add SafeAccount types and DTOs to safe.dto.ts
- Extend safe.hook.ts with SafeAccount state and API integration
- Support for selecting, creating, and updating SafeAccounts
- Permission-aware state (canWrite flag)

## Features
- SafeAccount list with selection support
- Automatic balance/history reload on SafeAccount change
- Legacy mode support (isLegacy flag)
- Access level tracking for permission-based UI

## Related
- Backend implementation: https://github.com/DFXswiss/api/pull/2602

## Test plan
- [ ] Verify SafeAccount list loads correctly
- [ ] Test SafeAccount selection updates balances
- [ ] Verify canWrite flag reflects access level
- [ ] Test legacy mode for users without SafeAccounts

### DIFF
--- a/src/dto/safe.dto.ts
+++ b/src/dto/safe.dto.ts
@@ -53,3 +53,39 @@ export interface CustodyHistoryEntry {
 export interface CustodyHistory {
   totalValue: CustodyHistoryEntry[];
 }
+
+// SafeAccount types
+
+export enum SafeAccessLevel {
+  READ = 'Read',
+  WRITE = 'Write',
+}
+
+export interface SafeAccountOwner {
+  id: number;
+}
+
+export interface SafeAccount {
+  id: number | null;
+  title: string;
+  description?: string;
+  isLegacy: boolean;
+  accessLevel: SafeAccessLevel;
+  owner?: SafeAccountOwner;
+}
+
+export interface SafeAccountAccess {
+  id: number;
+  userDataId: number;
+  accessLevel: SafeAccessLevel;
+}
+
+export interface CreateSafeAccountDto {
+  title: string;
+  description?: string;
+}
+
+export interface UpdateSafeAccountDto {
+  title?: string;
+  description?: string;
+}
--- a/src/hooks/safe.hook.ts
+++ b/src/hooks/safe.hook.ts
@@ -14,7 +14,17 @@ import {
 } from '@dfx.swiss/react';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { CustodyOrderHistory, CustodyOrderType, OrderPaymentInfo } from 'src/dto/order.dto';
-import { CustodyAsset, CustodyBalance, CustodyHistory, CustodyHistoryEntry } from 'src/dto/safe.dto';
+import {
+  CreateSafeAccountDto,
+  CustodyAsset,
+  CustodyBalance,
+  CustodyHistory,
+  CustodyHistoryEntry,
+  SafeAccessLevel,
+  SafeAccount,
+  SafeAccountAccess,
+  UpdateSafeAccountDto,
+} from 'src/dto/safe.dto';
 import { downloadPdfFromString } from 'src/util/utils';
 import { OrderFormData } from './order.hook';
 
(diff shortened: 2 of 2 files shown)


davidleomay (commented):
- [ ] Adapt to API changes
- [ ] Fix inexistent APIs?

TaprootFreak:
Heads-up before this goes further: the endpoint paths in `safe.hook.ts` do not exist on the API. The controller prefix is `custody/account`, not `safe-account`, so every call in this PR currently 404s.

| `safe.hook.ts` | API route |
| --- | --- |
| `safe-account` (GET, POST) | `custody/account` |
| `safe-account/${id}` (PUT) | `custody/account/${id}` |
| `safe-account/${id}/access` | `custody/account/${id}/access` |
| `safe-account/${id}/balance` | `custody/account/${id}/balance` |
| `safe-account/${id}/history` | `custody/account/${id}/history` |

Two of those did not exist at all until now:

- `.../balance` and `.../history` are being added in DFXswiss/api#4406, together with `.../order` and `.../pdf`. Until that lands there is no way to read *another* account's portfolio — the existing `custody`, `custody/history`, `custody/order` and `custody/pdf` endpoints always resolve to the calling user, which is exactly why the legacy fallback in this PR works and the account-scoped path does not.
- Granting, changing and revoking access comes with DFXswiss/api#4401. That PR also repairs the two guards behind `GET/PUT custody/account/:id`: they extend an abstract base holding the injected service but declared no constructor of their own, so no `design:paramtypes` metadata was emitted and Nest built them with no arguments. Both routes answered `403 Cannot read properties of undefined (reading 'checkAccess')` for every caller. Worth knowing if you tested against a deployed instance and concluded the account endpoints were simply denying access.

One more thing to align on: the legacy account. This PR models it with `isLegacy` plus a fallback to the old endpoints, which works. The API additionally accepts the literal id `legacy` on the account routes, and after #4401 that alias is only valid while the caller really has a legacy Safe — one custody user and no account row. Once the first grant materialises the Safe into a real account, the alias starts answering 404 and the numeric id must be used. If the hook caches account ids across that transition it will need to refetch.

Happy to push the path corrections onto this branch if you prefer — say the word and I will, otherwise it is a small change on your side once #4401 and #4402 are merged.

TaprootFreak:
Related: I have opened #1194, which implements the same feature against the endpoints as they exist today (`custody/account/*`, with the account-scoped read endpoints from DFXswiss/api#4406 and the access-level enforcement from DFXswiss/api#4417), including the picker component and the rule that a read-only account shows no transaction section.

It is not meant to push this one aside — happy to fold the work into this branch instead if you prefer. Whichever of the two we keep, the paths need to change from `safe-account/*` to `custody/account/*`.

TaprootFreak:
Closing in favour of #1194, which ships the same feature against the endpoints as they exist.

**Why this branch cannot land as it is.** It calls `safe-account/…` throughout — that path has never existed in the API. The endpoints are `custody/account/…`: the account list has been there since DFXswiss/api#4401, and the account-scoped reads for balance, history, orders and statement since DFXswiss/api#4406. So every request this branch makes would 404.

The second gap is that it changes only `safe.dto.ts` and `safe.hook.ts`. The hook exposes `selectSafeAccount`, but nothing renders a picker, so there is no way for a user to call it — the screen looks exactly as before.

**What #1194 does instead:** the picker component, the correct paths for balance, history, orders and statement, and one rule this branch did not cover — an account held by inspection only renders no deposit / withdraw / swap section at all. The API refuses those operations since DFXswiss/api#4417, so offering buttons that fail on click would be the wrong behaviour. It was exercised against a running instance with three accounts, one owned and two shared read-only.

**What was genuinely ahead here and is not lost:** the groundwork for creating and renaming accounts and for listing access grants. That belongs with the grant-management screens, which nobody has built yet — Rolf Schubiger can be given a read right today, but there is no screen on which he could hand one out. When that work starts, this branch is worth reading first; the shapes it sketched are close to right, only the paths need to move to `custody/account/…`.

Thanks — the direction was correct, the API simply landed somewhere else in the meantime.

PR #1089, opened 2026-05-12
feat(compliance): show phone call times in call-queue detail

bernd2022:
## Summary
- Move `phoneCallTimes` from `CallQueueAddressInfo` into `CallQueueUserInfo` (phone-call block), so it is visible in all five call-queue detail screens instead of only the IP-country variant where the address block is rendered.
- Adjust the address block title to `Address` (was `Address & Phone Call Times`).

## Test plan
- [ ] Open `/compliance/call-queues/UNAVAILABLE_SUSPICIOUS`, pick a user with `phoneCallTimes` set — value shows next to `Phone Call Status` in the right-hand User-Info column.
- [ ] Repeat for `MANUAL_CHECK_PHONE`, `MANUAL_CHECK_IP_PHONE`, `MANUAL_CHECK_IP_COUNTRY_PHONE`, `MANUAL_CHECK_EXTERNAL_ACCOUNT_PHONE`.
- [ ] User without `phoneCallTimes` still renders cleanly (shows `-`).


davidleomay (approved):
- [x] Fix merge conflicts

PR #1101, opened 2026-05-18
feat(support): add support notes tab and screen

bernd2022:
## Summary
- New compliance user-detail `NotesTab` with composer and list components
- Standalone `notes.screen` linked from the support dashboard for cross-user search/management
- Hooks for fetching, creating, updating, deleting notes and listing note users
- Routing entry in `App.tsx`

## Companion API
- DFXswiss/api: `feat/support-notes` — adds `/support/note` endpoints

## Test plan
- [ ] Open a user in compliance → Notes tab shows existing notes and allows create/edit/delete
- [ ] Visit support dashboard → "Notes" entry navigates to notes screen
- [ ] Filter / search by user on notes screen
- [ ] Verify role-based visibility (Support vs Compliance)

### DIFF
--- a/src/screens/notes.screen.tsx
+++ b/src/screens/notes.screen.tsx
@@ -8,7 +8,7 @@ import { SupportNoteInfo, SupportNoteScope, SupportNoteUser, useCompliance } fro
 import { useSupportDashboardGuard } from 'src/hooks/guard.hook';
 import { useLayoutOptions } from 'src/hooks/layout-config.hook';
 
-type Scope = Extract<SupportNoteScope, 'all' | 'free'>;
+type Scope = Extract<SupportNoteScope, 'All' | 'Free'>;
 
 export default function NotesScreen(): JSX.Element {
   useSupportDashboardGuard();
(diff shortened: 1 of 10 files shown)


src/screens/notes.screen.tsx
```diff
@@ -0,0 +1,264 @@
...
+      userDataId: selectedUser?.userDataId,
+    })
+      .then(setNotes)
+      .catch((e: unknown) => setError(e instanceof Error ? e.message : 'Failed to load notes'))
+      .finally(() => setIsLoading(false));
+  }, [listSupportNotes, submittedSearch, scope, selectedUser]);
+
+  useEffect(() => {
+    loadNotes();
+  }, [loadNotes]);
```
davidleomay, line 55: A single `refreshAll` in a `useEffect`?

davidleomay (approved):

src/screens/notes.screen.tsx
bernd2022: The two useEffects are intentionally separate. loadNotes depends on the filter state (search, scope, selectedUser) and re-runs on every filter change, while loadUsers only depends on the (stable) hook ref and therefore runs once on mount.
The customer list ("users with at least one note") doesn't change when filters change, so consolidating to a single refreshAll would refetch it on every keystroke / scope switch without any benefit. refreshAll is reserved for the cases where both actually need to update together — after create/edit/delete.

PR #1110, opened 2026-05-20
feat(support): add support issue templates (DEV-4571)

bernd2022:
## Summary
- New `/templates` screen and dashboard button to manage bilingual reply templates (DE required, EN optional)
- Template picker inside a support issue with placeholder syntax (`\$userData.firstname`, `\$transaction:last.id`, `\$issue.uid`, …), live preview, language auto-pick from the customer locale, and an array-selection dialog when multiple transactions match
- Send button now blocks while the message still contains unresolved placeholders
- Modal gained a `maxWidthClass` prop so the picker can render wider than the default dialog

Companion API PR: DFXswiss/api#3728

## Test plan
- [ ] Create / edit / delete templates from `/templates` (incl. bilingual content)
- [ ] Picker shows preview in customer language, falls back to DE if EN missing
- [ ] `\$transaction.id` opens the array picker when the customer has multiple transactions; `\$transaction:last.id` resolves directly
- [ ] Send button stays disabled while unresolved placeholders remain in the textarea
- [ ] Lazy-load of user data triggers only on first template-picker open per issue


src/components/support-templates/bilingual-content-editor.tsx
davidleomay, line 1: Rename `templates` folder to `support-templates` or similar

src/components/support-templates/template-list.tsx
```diff
@@ -0,0 +1,194 @@
...
+                    ✏️
+                  </button>
+                  <button
+                    type="button"
+                    className="px-2 py-0.5 text-xs text-dfxBlue-800 hover:bg-dfxRed-100/20 rounded transition-colors"
+                    onClick={() => setDeleteId(template.id)}
+                    title="Delete"
+                  >
+                    🗑️
+                  </button>
```
davidleomay: Use react icons instead?

src/hooks/support-templates.hook.ts
davidleomay, line 1: support-templates.hook.ts

src/screens/support-templates.screen.tsx
davidleomay, line 1: support-templates.screen.ts

davidleomay (approved):

davidleomay (approved):

PR #1137, opened 2026-06-30
feat(support): scope dashboard department filter & column by role

TaprootFreak:
## What

The support dashboard's **department filter** and **department column** were `admin`-only. They now show for any role that may see more than one department.

## Why

api PR DFXswiss/api#3983 made the **compliance** role a superset of support: compliance now sees Support *and* Compliance tickets in the same list. Without the filter/column, a compliance user gets a mixed list they can neither tell apart nor narrow. The backend already returns the correct set per role — this is purely the UI catching up.

## How

- New pure helper `visibleDepartmentsForRole(role)` in `src/util/support-helpers.ts`, mirroring the server-side `getVisibleDepartments` from #3983:
  - `support` → `[Support]`
  - `compliance` → `[Support, Compliance]`
  - `marketing` → `[Marketing]`
  - `admin` (no mapping) → all departments (unchanged)
- `support-dashboard.screen.tsx` shows the department filter + column when the role sees more than one department (`canFilterDepartment`) and scopes the filter options to that role's visible set. Admin behaviour is unchanged (all departments, filter + column visible).

## Scope & safety

- No API contract change and no change to what the server returns. The filter only ever sends a `department` value within the role's allowed set, and the server already ignores an out-of-set value (no escalation).
- `support` / `marketing` see a single department → no filter/column (unchanged).

## Tests

- New unit test `src/__tests__/support-helpers.test.ts` for `visibleDepartmentsForRole` (support / compliance-superset / marketing / admin-unrestricted / undefined-role).
- Full suite green locally: 27 suites, 288 tests. Prettier + ESLint clean.

Note: Playwright visual baselines intentionally do not run in CI (see CONTRIBUTING.md); not regenerated here since I don't have the local API stack running. Happy to add before/after screenshots if wanted.


TaprootFreak:
Verified locally against a full stack (api `develop` incl. DFXswiss/api#3983 + this branch) on macOS, signed in as a **COMPLIANCE** user, with open support issues seeded in both departments (3× Support, 2× Compliance):

- **Before** (develop): the dashboard shows only the Type + State filters and no department column — a compliance user sees the mixed Support/Compliance list but cannot tell the rows apart or narrow them.
- **After** (this branch): the **Department** filter (scoped to `[Support, Compliance]`) and the **Dept** column appear, so the mixed list is distinguishable and filterable.

Admin behaviour is unchanged (all departments, filter + column as before). Before/after screenshots are available on request.

PR #1141, opened 2026-06-30
fix(support): show issue ID, linkify URLs in messages, expand search hint

davidleomay:
## Problem

Three usability gaps in the Support Dashboard, reported in *Support Dashboard Refactoring 2*:

1. **Issue ID not shown** in the issue detail overview — only UID was displayed, even though support agents reference issues by their numeric ID.
2. **URLs in messages render as plain text** — a customer sending `https://...` in a chat message gets an unclickable string in the agent view.
3. **Search hint did not mention ID** — agents could not tell that searching by issue ID was meant to work.

The fourth original ask (preserve newlines in replies) is already fixed on develop via `whitespace-pre-wrap` on the message body.

## Fix

- **`support-dashboard-issue.screen.tsx`**:
  - Added `<InfoRow label=\"ID\" value={String(issueData.id)} mono />` at the top of the *Issue Details* panel.
  - Replaced the plain-text message render with a small `LinkedText` helper that splits the message on `https?://...` and wraps matches in `<a target=\"_blank\" rel=\"noopener noreferrer\">`. The `whitespace-pre-wrap` is kept, and `break-words` is added so long URLs don't overflow the bubble.
- **`support-dashboard.screen.tsx`**:
  - Updated the search placeholder to `\"Search by ID, UID, name, clerk, message...\"`.

## Backend follow-up

The placeholder now advertises ID search, but the backend `support/issue/list?query=…` only matches `name`, `uid`, `clerk`, userData fields, and message text — not `issue.id`. A companion PR in `DFXswiss/api` extends the existing OR-predicate with an `OR issue.id = :paramId` branch, gated on the term being fully numeric. Until that lands, typing a pure numeric ID into the search will not find the issue; this PR is intentionally frontend-only and consistent with the existing search pattern.

## Test plan

- [ ] Open any support issue → \"ID\" shows above \"UID\" in the Issue Details panel.
- [ ] Send a chat message containing `https://example.com/foo bar` → only the URL becomes a clickable link, the trailing `bar` stays plain.
- [ ] Send a message containing two URLs separated by text → both link, surrounding text unchanged.
- [ ] Confirm newlines are preserved in agent-side replies (already fixed on develop, regression check).
- [ ] Dashboard search field shows the updated placeholder.

### DIFF
--- a/src/screens/support-dashboard-issue.screen.tsx
+++ b/src/screens/support-dashboard-issue.screen.tsx
@@ -284,6 +284,7 @@ export default function SupportDashboardIssueScreen(): JSX.Element {
         {/* Info Panels - Row 1: Issue + Account */}
         <div className="flex gap-4 flex-wrap">
           <InfoPanel title="Issue Details">
+            <InfoRow label="ID" value={String(issueData.id)} mono />
             <InfoRow label="UID" value={issueData.uid} mono />
             <InfoRow label="Name" value={issueData.name} />
             <InfoRow label="Type" value={translate('screens/support', typeLabel(issueData.type))} />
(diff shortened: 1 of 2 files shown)


davidleomay:
⚠️ **Hold before merging — waiting on companion API change.**

The search placeholder in this PR now advertises ID search ("Search by ID, UID, name, clerk, message..."), but the backend `support/issue/list?query=…` does not yet match against `issue.id` — only `name`, `uid`, `clerk`, userData fields, and message text. If this merges before the API PR lands, the placeholder will be misleading: typing a pure numeric ID will return no results.

The companion API change extends the existing OR-predicate in `support-issue.service.ts` with an `OR issue.id = :paramId` branch, gated on the term being fully numeric.

Once that PR is merged and deployed, this one is safe to merge.

PR #1143, opened 2026-06-30
Remove Telegram support feature

TaprootFreak:
## What

Removes the Telegram support option from the app entirely:

- Deleted the Telegram support screen (`telegram-support.screen.tsx`, incl. the safety quiz)
- Removed the `/support/telegram` route and its lazy import in `App.tsx`
- Removed the Telegram tile (and `FaTelegram` icon) from the support screen
- Dropped the `/support/telegram` entry from the sitemap
- Removed `telegramDe` / `telegramEn` from `config/urls.ts`
- Removed the Telegram e2e test
- Removed the related `de` / `fr` / `it` translation keys

## Why

Telegram support is being discontinued. The support entry points now route users to **tickets** and **FAQ** only.

## Notes

- Old `/support/telegram` deep links degrade gracefully: react-router's top-level `errorElement` renders the standard error screen (no white-screen).
- The generic `Contact us` translation key is now unused but left untouched on purpose (it is a shared label, not Telegram-specific).

## Checks (local)

- `tsc --noEmit` — clean (no errors in changed scope)
- `npm run lint` — clean
- `npm run build` (prod) — succeeds
- All three translation JSON files parse as valid JSON


TaprootFreak:
Heads-up for reviewers: this changes the **Support** screen (the Telegram tile is removed), so the local visual baseline `e2e/screenshots/baseline/user-flows.spec.ts-support-main-chromium-darwin.png` is now stale (it still shows the Telegram tile). These visual tests intentionally do not run in CI (local/code-review aid per CONTRIBUTING), so this does not block the PR. The baseline should be regenerated locally against the seeded stack (`npx playwright test user-flows -g "support" --update-snapshots`) and committed as a follow-up.

PR #1164, opened 2026-07-04
feat(dashboard): add admin page to set financial log validity

TaprootFreak:
## What
New admin-only screen at `/dashboard/financial/log-validity` to manually set the `valid` flag on log entries.

Two independent forms:
- **By log ID** — set a single entry's validity via `PUT /log/:id`.
- **By financial range / threshold** — bulk-update FinancialDataLog entries via `PUT /log/financial/validity`, filtering by `from`/`to` (created window, from inclusive / to exclusive) and/or `min`/`max` (exclusive on `totalBalanceChf`). Reports the number of affected rows.

## Why
FinancialDataLog balance snapshots occasionally show transient outliers (e.g. a large deposit whose plus and minus legs land in different one-minute snapshots). The automatic validator flags these, but some cases need manual review. This page lets an admin correct the `valid` flag directly from the dashboard.

## Notes
- Gated with `useAdminGuard` — uses the ADMIN role the backend endpoints already require.
- Client-side validation mirrors the backend rules (at least one filter; `from <= to`; `min < max`). Date inputs require an explicit timezone and are normalized to UTC ISO before sending, matching the UTC timestamps of the financial data log.
- Frontend-only; both endpoints already exist in the API.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -99,6 +99,7 @@ const DashboardFinancialHistoryScreen = lazy(() => import('./screens/dashboard-f
 const DashboardFinancialLiveScreen = lazy(() => import('./screens/dashboard-financial-live.screen'));
 const DashboardFinancialExpensesScreen = lazy(() => import('./screens/dashboard-financial-expenses.screen'));
 const DashboardFinancialLiquidityScreen = lazy(() => import('./screens/dashboard-financial-liquidity.screen'));
+const DashboardFinancialLogValidityScreen = lazy(() => import('./screens/dashboard-financial-log-validity.screen'));
 const SitemapScreen = lazy(() => import('./screens/sitemap.screen'));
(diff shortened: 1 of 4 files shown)


TaprootFreak:
Ready for review — this went through 3 review passes before it came back clean.

Addressed along the way:
- Made the page discoverable: added it to the Financial Dashboard index grid and to the sitemap, alongside its siblings.
- Added a confirmation step before applying a validity change — it echoes the target log id / resolved filters and the new value, since the bulk range update can affect many rows.
- Switched the from/to filters to a native date+time picker (empty by default; the local selection is sent as a UTC ISO instant).
- Interpolate the validated log id directly in the request URL (dropped a redundant numeric conversion).

CI is green (lint, tests, build:dev).

PR #1171, opened 2026-07-08
feat(realunit): show all compliance customers upfront

Blume1977:
## Summary

The RealUnit compliance dashboard (`/realunit/compliance`) now shows **all** customers with a registered RealUnit wallet when opened; the search only narrows the list down. An empty search shows the full list again.

- `useRealunitCompliance.searchCustomers`: `key` is optional — without it the api returns the complete tenant scope
- Screen loads the full list on mount; the Search button is only disabled while loading
- e2e spec: asserts the upfront list is rendered before any input

## Depends on

API pair PR: DFXswiss/api#4135 — without it, the initial no-key request is rejected (`key` was `@IsNotEmpty`). Merge the api PR first.

## Visual baselines

Not regenerated: the committed `01-search` screenshot is taken after a search and is visually unchanged by this PR. The new upfront-list state has no committed baseline yet; can be added once the local stack is up.

## Tests

- All 32 suites green (369 tests), ESLint clean

### DIFF
--- a/src/hooks/realunit-compliance.hook.ts
+++ b/src/hooks/realunit-compliance.hook.ts
@@ -1,9 +1,11 @@
+import { ResponseType } from '@dfx.swiss/react';
 import { useMemo } from 'react';
 import {
   RealUnitCustomerDetailDto,
   RealUnitCustomerListDto,
   RealUnitKycFileDownloadDto,
 } from 'src/dto/realunit-compliance.dto';
+import { downloadFile as saveFile, filenameDateFormat } from 'src/util/utils';
 import { useGuardedApi } from './guarded-api.hook';
 
 // RealUnit tenant compliance hook. READ-ONLY, strictly customer-scoped `/v1/realunit/compliance/*` endpoints over
@@ -34,11 +36,23 @@ export function useRealunitCompliance() {
     });
   }
 
+  // Full dossier as ZIP (all allowlisted files, bundled and audit-logged api-side).
+  async function downloadDossier(id: number): Promise<void> {
+    const { data, headers } = await call<{ data: Blob; headers: Record<string, string> }>({
+      url: `realunit/compliance/customers/${id}/dossier`,
+      method: 'GET',
+      responseType: ResponseType.BLOB,
+    });
+
+    saveFile(data, headers, `RealUnit_dossier_${id}_${filenameDateFormat()}.zip`);
+  }
+
   return useMemo(
     () => ({
       searchCustomers,
       getCustomer,
       downloadFile,
+      downloadDossier,
     }),
     [call],
   );
--- a/src/hooks/realunit-compliance.hook.ts
+++ b/src/hooks/realunit-compliance.hook.ts
@@ -12,9 +12,10 @@ import { useGuardedApi } from './guarded-api.hook';
 export function useRealunitCompliance() {
   const { call } = useGuardedApi();
 
-  async function searchCustomers(key: string): Promise<RealUnitCustomerListDto[]> {
+  // Without a key the api returns the complete tenant scope (all RealUnit customers).
+  async function searchCustomers(key?: string): Promise<RealUnitCustomerListDto[]> {
     return call<RealUnitCustomerListDto[]>({
-      url: `realunit/compliance/customers?key=${encodeURIComponent(key)}`,
+      url: `realunit/compliance/customers${key ? `?key=${encodeURIComponent(key)}` : ''}`,
       method: 'GET',
     });
   }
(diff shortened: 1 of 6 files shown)


src/screens/realunit-compliance.screen.tsx
```diff
@@ -26,17 +26,23 @@ export default function RealunitComplianceScreen(): JSX.Element {
...
     setResults(undefined);
-    searchCustomers(searchKey.trim())
+    searchCustomers(key)
       .then((res) => setResults(res))
       .catch((e: Error) => setError(e.message ?? 'Unknown error'))
       .finally(() => setIsLoading(false));
   }
 
+  function handleSearch(): void {
+    loadCustomers(searchKey.trim() || undefined);
```
TaprootFreak, line 43: The Enter path isn't gated on `isLoading`, unlike the Search button (`disabled={isLoading}` on line 62). With the new mount effect always firing a keyless full-list request, typing a key and pressing Enter before that request resolves puts two requests in flight. Both call `setResults` in `.then`, so the rendered table is last-write-wins by resolution order: the small search result can render first and then be overwritten by the full list, leaving every customer shown while the search box still holds the key.

Minor (staff-only, read-only, self-corrects on the next search), but newly guaranteed now that the mount load is always in flight. Suggest gating `handleSearch` on `!isLoading` (mirroring the button), or discarding stale responses via a request token.

src/hooks/realunit-compliance.hook.ts
```diff
@@ -12,9 +12,10 @@ import { useGuardedApi } from './guarded-api.hook';
 export function useRealunitCompliance() {
   const { call } = useGuardedApi();
 
-  async function searchCustomers(key: string): Promise<RealUnitCustomerListDto[]> {
+  // Without a key the api returns the complete tenant scope (all RealUnit customers).
+  async function searchCustomers(key?: string): Promise<RealUnitCustomerListDto[]> {
     return call<RealUnitCustomerListDto[]>({
-      url: `realunit/compliance/customers?key=${encodeURIComponent(key)}`,
+      url: `realunit/compliance/customers${key ? `?key=${encodeURIComponent(key)}` : ''}`,
```
TaprootFreak, line 20: This keyless branch (URL without `?key=`) is the core new behavior of the PR but has no unit test: the only `searchCustomers` test asserts the keyed URL, and the e2e spec that exercises the keyless path doesn't run in CI (per CONTRIBUTING). The sibling `useRealunitSupport` in the same test file already has an `omits the query string when no list params are given` case — mirroring it here would stop a keyless-only regression (e.g. emitting `?key=undefined`) from passing green CI.

One-liner: `searchCustomers()` → expect url `realunit/compliance/customers`.

TaprootFreak (commented):
Code review — 2 minor points, no blocker; both are one-liners. Details inline. (Reminder: this depends on DFXswiss/api#4135, so api must be deployed first, otherwise the keyless mount request is rejected.)

Blume1977:
Follow-up commit 3c381fe: zwingend sichtbares **Checks**-Panel im Customer-Detail (Ident-Check Sumsub + Dilisense-Check, fehlend = rotes Missing-Badge) — Pair-PR dafür ist DFXswiss/api#4140 (NAME_CHECK-Allowlist + REALU/ZCHF-Transaktionsfilter). API-PR zuerst mergen.

Blume1977:
@TaprootFreak Ready for review — Frontend-Gegenstück zu DFXswiss/api#4140 (bitte api-PR zuerst mergen).

Blume1977:
@TaprootFreak Beide Punkte umgesetzt in 35216ac: Enter-Pfad ist jetzt wie der Search-Button auf `isLoading` gegated (kein Race zwischen Mount-Full-List und schneller Suche mehr), und der keyless Full-List-URL-Fall ist als Unit-Test gepinnt (`searchCustomers()` → `realunit/compliance/customers` ohne `?key=`).

TaprootFreak:
Addressed the open points on this PR:

- **Fixed the widget build** (`Build and test`), which was red on `@typescript-eslint/no-non-null-assertion` for the ident/name-check download button (`check.fileUid!` / `check.fileName!`). The values are now captured as narrowed `const` locals so the download closure keeps their type without an assertion; the widget build compiles again.
- Added a unit test for the new `downloadDossier` blob request in the compliance hook.
- Restored the e2e search test's discriminating power: it now waits for the keyed request on `Enter`, since the upfront keyless list load otherwise satisfied the post-`Enter` assertions on its own (a broken search wiring would have stayed green).

Completed three review passes (conformity + logic) to zero findings. Lint, the full test suite (32 suites / 371 tests) and the widget build are green.

Note: the `review` check is red only because a fork PR's `GITHUB_TOKEN` is read-only and the bot cannot post its summary comment (`403`); the checks it runs (commit signatures, ESLint, TypeScript) pass. The listed unverified commits are the fork's original commits — not a merge blocker.

PR #1173, opened 2026-07-09
feat(mail): redirect merged accounts to master code on mail updates

TaprootFreak:
## What
Companion to DFXswiss/api#4092. That PR makes the mail-update routes return HTTP 401 + `MergedDto` (carrying `switchToCode`) for merged accounts instead of a bare 409. This handles that response on the mail flows so a merged account is redirected to its master account's code — the same behaviour `kyc.screen` already has — instead of hitting a generic session error / logout.

## Changes
- Bump `@dfx.swiss/react` to `^1.5.0` (adds `switchToCode` on `ApiException`, from DFXswiss/packages#187).
- New `useMergedAccount` hook (`handleMergedError`): on a 401 carrying `switchToCode`, navigate to `/kyc?code=<master>` + logout.
- Apply it as the first `catch` branch in `edit-mail.screen` (`onSubmit`/`onVerify`) and `mail.edit` (`saveUser`).

## Notes
- The redirect targets `pathname: '/kyc'` explicitly (not just `?code=` like `kyc.screen`, which works only because that screen already lives on `/kyc` and reads the param) — so the master code takes effect from the mail screens too.
- Redirect + logout only; no new UI / i18n.
- Prerequisite packages change is published as `@dfx.swiss/react@1.5.0`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.4",
       "license": "MIT",
       "dependencies": {
-        "@dfx.swiss/react": "^1.4.1",
+        "@dfx.swiss/react": "^1.5.0",
         "@dfx.swiss/react-components": "^1.3.2",
         "@ledgerhq/hw-app-btc": "^6.24.1",
         "@ledgerhq/hw-app-eth": "^6.33.7",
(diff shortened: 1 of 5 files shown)


TaprootFreak:
### Review summary

Reviewed (correctness + conventions), no findings:

- `useMergedAccount` narrowly matches only a 401 that carries `switchToCode`; it redirects to `/kyc?code=<master>` and logs out. The explicit `/kyc` pathname is required here — `kyc.screen`'s bare `?code=` works only because that screen already lives on `/kyc` and reads the param — and the `?code` deliberately survives the subsequent `logout()` (keeps `kycCode` truthy, so the login guard doesn't fire).
- Wired as the first `catch` branch in `edit-mail.screen` (`onSubmit`/`onVerify`) and `mail.edit` (`saveUser`, with `throw e` preserving prior behaviour). The existing 409 `exists`/`merge` and 403 handling stays intact; only the merged-account case changes.
- `@dfx.swiss/react ^1.5.0` + lock (react 1.5.0 / core 0.4.0) consistent — `Build and test` (npm ci + tsc + tests) green. No conflict with `@dfx.swiss/react-components` (no core dependency).
- Redirect + logout only — no new UI / i18n, so the custom Tailwind palette isn't touched.

Merge together with DFXswiss/api#4092.

PR #587, opened 2025-05-19
[DEV-3971] fix env variable + refactor url constructions

lapatric:

PR #1178, opened 2026-07-15
fix(referral): stop forcing a trailing slash onto the copied referral link

Blume1977:
## Problem

Since the URL-construction refactor in #587 (May 2025), the account screen's referral copy button produces `https://dfx.swiss/app/services/?code=XXX-XXX` — with a trailing slash before the `?`. The `url()` helper normalizes the base to end with a slash for path joining, and with an empty path the slash leaks into the result.

dfx.swiss only 301-redirects the slash-less form to `api.dfx.swiss/v1/app/services?code=…` (where the ref is stored per IP); the slash form returns the static marketing page (HTTP 200) and the code is lost — no meta refresh, no script forwards it, and the app itself only reads `refcode`/`special-code`/`recommendation-code`. Every freshly copied referral link therefore silently loses its attribution.

## Fix

With no path to join, `url()` now uses the base exactly as provided (`new URL(base)`); path joining is unchanged. Covered by new `url()` unit tests (base kept verbatim with/without slash, joining, absolute-path-as-base).

## Notes

- Ops side: a redirect rule for the slash variant on dfx.swiss would additionally heal all already-shared links (this PR only fixes newly copied ones).
- Backend counterpart for the related mail-invitation ref gap: DFXswiss/api#4211 (independent bug, no pairing required).
- Full test suite green (33 suites / 384 tests).

### DIFF
--- a/src/__tests__/utils.test.ts
+++ b/src/__tests__/utils.test.ts
@@ -38,6 +38,7 @@ import {
   formatLocationAddress,
   apiUrl,
   relativeUrl,
+  url,
 } from '../util/utils';
 
 describe('utils', () => {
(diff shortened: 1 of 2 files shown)


Blume1977:
@Yannick1712 Review-Request: Frontend-Teil zum Ref-Link-Fix (Backend-Gegenstück: DFXswiss/api#4211).

PR #1179, opened 2026-07-16
feat(realunit): default-hide empty accounts in compliance customer list

Blume1977:
## Problem

Since api#4212 the RealUnit compliance customer list is complete by design: every account with a RealUnit wallet address appears, including accounts that only created a wallet (no name, no mail, KYC 0). Correct for compliance, but the default view is cluttered with rows showing nothing but an ID.

## Changes

One presentation-only filter; the API keeps returning the complete list and **no account can disappear**:

- A single default-on toggle above the table: **"Hide empty accounts (n)"** with a live count, next to the total ("Customers: N"). One click reveals the hidden rows — in their original API position, the list is never reordered.
- **"Empty"** = no name, no mail, no KYC progress in either direction (a terminated/rejected account has a non-zero level and stays visible), and a **resolved** zero balance. An unresolved balance (`undefined`, e.g. indexer outage) fails open to visible, so an infrastructure failure can never hide a real REALU holder. Likewise a missing or blank `kycLevel` is treated as unresolved and fails open to visible; only a resolved level of 0 counts as no KYC progress.
- **An active search bypasses the filter** and always shows every match; the toggle is not rendered during a search. Whoever looks for a specific customer must find them — that blind spot is exactly what this screen suffered from before api#4212.
- When the filter hides every loaded row, the empty state says so explicitly ("All accounts are hidden by the filter above") instead of the generic "No entries found".

A previous revision of this PR also shipped a hardcoded internal test-account mail-pattern filter; it was dropped after review (personal mail address in the public bundle, client-side business rule). Flagging test accounts is a data topic (DB/API), not a rendering one.

## Tests

- Unit suite for `isEmptyAccount` covering the fail-open unresolved-balance case and negative KYC levels.
- Full unit suite passes (34 suites, 388 tests); lint and Prettier clean.
- Playwright baseline `realunit-compliance-01-search.png` regenerated on macOS against the local api stack and committed with this change.

## Review follow-ups (maintainer)

- "Empty" additionally requires an untouched `kycStatus` (`'NA'`): a level-0 account with a compliance-relevant status (e.g. Check/Rejected/Terminated) can never be hidden.
- The hide-empty toggle only renders when there is at least one empty account to hide.
- The screen wiring (search bypass, toggle, empty-state messages) is now covered by component tests; filter edge cases (numeric wire-format `kycLevel`, empty/non-numeric strings, fractional/negative balances) are pinned in the unit suite.
- The e2e fixtures mirror the `balance` field and include an empty account (hidden by default, visible in search) and drive the regenerated search baseline.


Blume1977:
@TaprootFreak Review-Request: Follow-up zur Compliance-Kundenliste nach api#4212 (leere + interne Testaccounts standardmässig ausgeblendet, mit Zähler und Toggle; Suche zeigt immer alles).

Blume1977:
Nach Review vereinfacht (8eebd66): nur noch EIN Toggle "Leere Accounts ausblenden" mit Zähler; Testaccount-Pattern-Filter entfernt (private Mail-Adresse gehört nicht ins öffentliche Bundle, Klassifikation wäre ein API/DB-Thema); unresolved Balance zählt jetzt als sichtbar (fail-open, Indexer-Ausfall kann keine Aktionäre mehr verstecken); keine Umsortierung mehr (API-Reihenfolge bleibt); filterbewusster Empty-State; Toggle bei aktiver Suche ausgeblendet statt disabled.

TaprootFreak:
Took 4 full review passes (conformity + logic, both lenses each pass) to reach zero findings. Addressed along the way: title/body brought in line with the actual diff; the empty filter additionally requires an untouched kycStatus and fails open on missing, null or blank kycLevel; component tests for the screen wiring (search bypass, toggle, empty-state messages, filter re-engagement after a cleared search); pinned filter edge cases (numeric wire-format level, fractional/negative balances); the toggle only renders when there is something to hide; e2e fixtures now include an empty account with default-hide and search-bypass assertions; regenerated the 01-search baseline on macOS. All checks green on fbd2593.

PR #1182, opened 2026-07-21
perf(wallet): port MetaMask hook off web3 to viem

Danswar:
## Summary

- `src/hooks/wallets/metamask.hook.ts`: replaced all `web3.eth.*` / `Contract.methods.*` calls with `viem`'s `publicClient`/`walletClient` (`getBalance`, `readContract`, `writeContract`, `signMessage`, `sendTransaction`, `getAddresses`), mirroring the pattern already used in `wallet-connect.hook.ts`.
- `src/hooks/web3.hook.ts`: the one remaining `web3.utils.toHex` call replaced with viem's `numberToHex`.
- Removed `web3` from `package.json` (the only one of the three that was ever a direct dependency — `web3-core`/`web3-eth-contract` were transitive/type-only imports resolved through it) — confirmed no other file in the repo imports any of them.
- Added a `TextEncoder`/`TextDecoder` polyfill to `src/setupTests.ts`: jsdom's test environment doesn't provide these, and viem needs them at import time. This was latent — `wallet-connect.hook.ts` already used viem but had no test coverage, so nothing had hit this before.
- Removed the now-unused `jest.mock('web3', ...)` blocks from the two existing test files, and added real unit test coverage for `readBalance`, `createTransaction`, and `sign` against mocked viem clients (this logic had zero coverage before this PR).

### Why

`web3` is CommonJS-only with no `sideEffects: false` declaration (confirmed via its published `package.json`), so it cannot be tree-shaken — importing any part of it always ships the whole package, plus its own legacy crypto-polyfill dependency tree (`elliptic`, `bn.js`, `eth-lib`, `secp256k1`, `asn1.js`, IPFS-era `multihashes`/`multibase` for the long-dead `web3-bzz` Swarm support, etc.). Since `useMetaMask` is wired into the app-wide `WalletContextProvider` (not behind a lazy route), all of that shipped in the main entry chunk for every visitor, regardless of wallet choice.

### Measured impact

Same-checkout before/after production build:

| | `main.js` |
|---|---|
| Before (`web3`) | 3,583,689 bytes (3.42 MiB) |
| After (`viem`) | 2,516,089 bytes (2.40 MiB) |
| **Cut** | **1,067,600 bytes (1.02 MiB) — 29.8%** |

### Behavior preserved intentionally

- The original code's `.send()`/`.sendTransaction()` blocks until the transaction is *mined* before returning the hash (web3's `PromiEvent` resolves with the receipt). `viem`'s `writeContract`/`sendTransaction` return the hash immediately instead, so an explicit `publicClient.waitForTransactionReceipt({ hash })` was added to keep today's wait-for-mined behavior unchanged. Note this means MetaMask and WalletConnect (`wallet-connect.hook.ts`, which does *not* wait) already behave differently here — out of scope for this PR, flagging for awareness.
- The original code explicitly nulled `maxFeePerGas`/`maxPriorityFeePerGas` on every send (added in #163/DEV-2129, for MetaMask/chain combinations that misbehaved with EIP-1559 fee fields). The first version of this port dropped that whenever no `gasPrice` override was given, silently reverting to viem's default fee estimation. Fixed in a follow-up commit: `createTransaction` now resolves `gasPrice` via `publicClient.getGasPrice()` when no override is passed, so `gasPrice` is always a concrete value and the EIP-1559 fields are never set. Note this restores the same *request shape* as before at the code level, but — same as the rest of this PR — hasn't been independently verified against a live wallet.
- `readBalance()`'s native-coin branch had a pre-existing bug carried over from the web3 code (`return promise.then(...)` inside a `try` doesn't get caught by that `try`'s own `catch`), which the ERC20 branch didn't have. Fixed to match.

## Test plan

- [x] `npm test` — all 431 tests pass (36 suites), including new coverage for `readBalance`/`createTransaction`/`sign`
- [x] `tsc --noEmit` clean on all non-test application code
- [x] Local production build succeeds; `main.js` measured before/after in the same checkout
- [ ] Manual QA against a live MetaMask wallet (balance reads, native + ERC20 transfer, `personal_sign`, and confirming the broadcast transaction is actually type `0x0`/legacy) on testnet before merging — not done as part of this PR


Danswar:
This PR went through 3 review passes before reaching 0 outstanding issues.

Fixed along the way:
- Restored the legacy gas-pricing behavior (see #163/DEV-2129) that the initial port had silently dropped for transactions without an explicit gas price override.
- Fixed a pre-existing bug in the native-coin balance read where a rejected balance fetch wasn't being caught by its own error handling.
- Added unit test coverage for the balance-read, transaction-construction, and signing logic, which previously had none.

Danswar:
## Manual QA checklist before merge

This PR hasn't been tested against a live wallet yet (see test plan). Checklist for whoever picks that up:

- [ ] Connect/disconnect MetaMask; confirm wallet detection works (and Rabby/Trust/Coinbase in-app-browser variants if reachable)
- [ ] Read native-coin balance — compare against the wallet's own displayed balance
- [ ] Read an ERC20 balance (e.g. USDT/USDC/dEURO) — compare against the wallet's own displayed balance
- [ ] Send a native-coin transaction end-to-end (Buy/Sell/Swap) — confirm it broadcasts and the returned hash matches what actually lands on-chain
- [ ] Send an ERC20 transfer end-to-end — confirm the transferred amount is correct (decimals scaling) and the hash matches
- [ ] Confirm the broadcast transaction is actually type `0x0` (legacy), not EIP-1559 — check via a block explorer or `eth_getTransactionByHash`. This is the one behavior this PR restores by code shape only and hasn't been verified against real wallet behavior.
- [ ] `personal_sign` message signing — confirm the signature validates server-side (whatever auth/session flow consumes it)
- [ ] "Add token to MetaMask" (`wallet_watchAsset`) for an ERC20 asset — confirm symbol/decimals show up correctly in the wallet
- [ ] Network/chain-switch flow across a couple of supported chains, including the "add chain" fallback for one not yet added to the wallet
- [ ] Payment-link "Pay with MetaMask" flow specifically (uses the gasPrice override path) — confirm it still works with the resolved-gasPrice change

Danswar:
Superseded by #1330 — same change rebased onto current develop, with both touched hooks brought to full coverage per the new CONTRIBUTING rules. This branch lives in the base repo and can no longer be updated, so the continuation is opened from the fork.

PR #1186, opened 2026-07-22
fix: reload once on ChunkLoadError (recover stale chunks after deploy)

TaprootFreak:
## What

Adds a global `ChunkLoadError` handler in `src/index.tsx` that reloads the page **once** (loop-guarded) when a stale/missing chunk fails to load.

## Why

After a deploy, the content-hashed chunks are replaced. A tab left open across a deploy can request a chunk that no longer exists. On a static host that serves `index.html` (200) for unknown paths (e.g. Cloudflare Pages via the SPA fallback), this surfaces as a **`ChunkLoadError`** (HTML parsed as JS) rather than a clean 404 — the app would otherwise stay broken until a manual reload.

## How

- `isChunkLoadError(message)` — regex match for JS/CSS chunk-load failures.
- `reloadOnceForChunkError()` — `window.location.reload()`, guarded by a `sessionStorage` timestamp (`dfx.chunkReloadAt`, 10s) to prevent a reload loop.
- Listeners on `error` and `unhandledrejection` (the latter casts `event.reason as Error | undefined` — no `any`).
- Inserted before `ReactDOM.createRoot`; existing session-clearing / `root.render` untouched.

## Verification

Verified on Node 20 (CI pin): `npm ci`, `npx eslint src/index.tsx` clean, `npm run build:dev` compiles successfully; `tsc` error count identical before/after (no new errors from `index.tsx`).

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -22,13 +22,23 @@ if ((urlParams.has('address') && urlParams.has('signature')) || urlParams.has('s
 function isChunkLoadError(message?: string): boolean {
   return !!message && /Loading chunk [\w-]+ failed|ChunkLoadError|Loading CSS chunk [\w-]+ failed/i.test(message);
 }
+// A new deploy replaces the content-hashed chunks. A tab left open across a deploy can
+// request a chunk that no longer exists; the static host serves index.html (200) for it,
+// which surfaces as a ChunkLoadError. Reload once to pick up the new chunks. The guard
+// lives in localStorage (it survives the sessionStorage.clear() above that runs on
+// session/login URLs) and is time-boxed, so a persistent failure reloads at most once per
+// window instead of looping. Storage access is wrapped because embedded/iframe contexts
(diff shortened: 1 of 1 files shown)


TaprootFreak:
Reviewed across 2 rounds. Round 1 surfaced a high-severity loop: the reload guard lived in `sessionStorage`, but `index.tsx` calls `sessionStorage.clear()` on session/login URLs (`address`+`signature` or `session`), which wiped the guard on every reload — a persistent chunk error on such an entry point would loop unbounded. Fixed by moving the guard to `localStorage` (survives the clear), wrapping storage access in try/catch (embedded/iframe contexts can block it — skip the reload then), and widening the window to 30s (at most one reload per window). Round 2 verified the fix — 0 findings. Known low-severity edge: a future-dated client clock could persistently disable the auto-reload until storage is cleared (non-blocking). CI green.

PR #1188, opened 2026-07-23
List RealUnit on the payment-link page (ZCHF on Ethereum)

TaprootFreak:
RealUnit settles OpenCryptoPay payments on-chain in ZCHF on Ethereum (it consumes the LNURL only as the payment-request identifier), so it is **not** a Lightning wallet. Two changes let a correctly-classified non-Lightning wallet work on `/pl`:

- `getDeeplinkByWalletId`: a RealUnit case that builds the payload-bearing deeplink (`<deepLink>lightning:<LNURL>`) from the backend `deepLink` prefix and the LNURL — the generic path only injects the payload for Lightning wallets.
- `hasActionDeepLink`: treat RealUnit as having an action deeplink so the UI shows "Pay in app" instead of "Open app and scan QR code again".

`Wallet.qualifiesForPayment` already matches a ZCHF/Ethereum wallet (via `supportedAsset.name === 'ZCHF'`); tests added for that plus the deeplink/label behaviour.

Backend companion (the `wallet_app` row): DFXswiss/api#4339.

### DIFF
--- a/src/__tests__/payment-link-wallet.test.ts
+++ b/src/__tests__/payment-link-wallet.test.ts
@@ -145,5 +145,21 @@ describe('Wallet', () => {
       const result = Wallet.qualifiesForPayment(wallet, transferInfoList);
       expect(result).toBe(true);
     });
+
+    it('should return true for RealUnit-like Ethereum/ZCHF wallet against Ethereum transfer', () => {
+      const wallet = createWallet(['Ethereum'], [{ name: 'ZCHF', uniqueName: 'Ethereum:ZCHF' }]);
+      const transferInfoList = [createTransferInfo('Ethereum', ['ZCHF'])];
+
+      const result = Wallet.qualifiesForPayment(wallet, transferInfoList);
+      expect(result).toBe(true);
+    });
+
+    it('should return false for RealUnit-like Ethereum/ZCHF wallet against Lightning-only transfer', () => {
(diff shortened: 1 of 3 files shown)


TaprootFreak:
Reviewed across conformity and logic (independent passes) with no findings; `Build and test` (lint + test + build:dev + widget:dev) is green.

Adds a RealUnit case to `getDeeplinkByWalletId` that builds the payload-bearing `<deepLink>lightning:<LNURL>` deeplink (the generic path only injects the payload for Lightning wallets), and extends `hasActionDeepLink` so the UI shows "Pay in app". `Wallet.qualifiesForPayment` already matches a ZCHF/Ethereum wallet via `supportedAsset.name === 'ZCHF'`; covered by the added `qualifiesForPayment` cases plus a `renderHook` test asserting the exact deeplink string and label.

Backend companion (the `wallet_app` row): DFXswiss/api#4339.

TaprootFreak:
Reworked to be config-driven and reviewed (conformity + logic) to zero findings. There is now **no wallet name** in the deeplink/label logic — the LNURL action-deeplink + "Pay in app" are gated on the backend `wallet_app.hasActionDeepLink` flag, additive to the existing Lightning check, so any future non-Lightning OpenCryptoPay wallet needs only its backend row (no frontend code). Every existing wallet's behaviour is unchanged (Lightning wallets append; bare-open wallets like Cake Wallet/urble stay bare). Tests cover the flag-driven, the bare-open, and the Lightning paths (the last a non-vacuous regression guard). Companion: DFXswiss/api#4352 sets the flag for RealUnit.

PR #1190, opened 2026-07-24
Add Bank Frick personal IBAN buy parameter

TaprootFreak:
## Summary

- support `/buy?asset-in=EUR&personal-iban=frick`
- expose the React prop as `personalIban` and the Web Component attribute as `personal-iban`, mapping `frick` case-insensitively to the API enum value
- derive the selector from its source on every read — the URL search in the browser, the live widget property when embedded — instead of copying it into component state
- ask the customer to confirm before a personal IBAN is requested, and verify that the response actually honoured the request
- display the bank returned by the API together with the personal IBAN payment details
- map the API's Frick error tokens to translated, cause-specific messages
- document the public URL and component contracts

## Why the selector has no state

The first implementation kept the selector in component state. Every subsequent way of storing it
turned out to have a path on which state and source drift apart, and on each of those the customer
requests a personal IBAN and silently receives the default bank: not persisted at all (lost on every
reload), `localStorage` (never cleared, so it applied to later unrelated purchases), `sessionStorage`
with logout detection (plain token expiry is indistinguishable), an in-memory copy (browser back, a
cancelled sell flow, a second purchase in a persistently mounted widget).

Deriving the value removes the class of bug rather than its instances: no copy can drift, there is no
lifecycle to keep in sync, and no clearing can fire from an unrelated flow. The customer also sees
the selector in the address bar for as long as it applies.

## Why the customer is asked

A personal IBAN causes a real account to be created at the bank, and that cannot be undone. So the
app asks before the first quote that would carry the selector, once per tab and authenticated
customer, held in memory for the running instance — a reload asks again, which is proportionate for
an irreversible action. Declining requests ordinary details instead; that decision applies to the
purchase at hand, and a selector arriving anew asks again.

The alternative was to decide it silently, and five rounds of trying showed why that does not work:
the same URL can mean a leftover from the previous customer or a deliberate request by the current
one, and no amount of session-state detection can tell them apart. Asking is the honest answer to a
genuinely ambiguous case.

## Who the account belongs to

CHF accounts at Yapeal are genuinely held by the customer; EUR accounts at Bank Frick are not — the
account stays with DFX and the customer gets an IBAN assigned to them alone. Every customer-facing
string was checked against that: the confirmation states it, the Yapeal wording is unchanged because
it is true there, and a test asserts the rendered prompt per language rather than its exact wording.

A missing currency on the self-service screen is now a visible error instead of a silent EUR default,
and the Bank Frick branch on that screen was removed — its endpoint only knows Yapeal, so the branch
was unreachable.

## Response verification

An explicit Frick request is checked against the response: it must be marked as a personal IBAN, come
from Bank Frick, and name DFX AG. Against an older API the request field is stripped, so without this
check the customer would silently receive ordinary bank details — and the legacy path names the
customer as recipient, which is exactly the payee mismatch this change exists to prevent. On a
mismatch the details are withheld behind an explicit choice, and choosing the plain alternative
fetches fresh selector-free details rather than showing the rejected response.

## Error handling

The API returns short tokens for this feature, mapped to translated messages in de/fr/it and split
into two scopes: the buy flow, and the reconstruction of stored payment details for invoices and
receipts. The feature-specific mapping applies **only** when the request actually carried the
provider — it previously also caught ordinary EUR bank transfers and replaced the generic KYC dialog
(with its working action) by an unusable retry hint.

## Fixes beyond the feature

- `relativeUrl` merges an existing query string instead of appending a second `?`. The external
  magic-link and Alby return targets were path-only and dropped **all** query parameters.
- Overlapping quote fetches can no longer let a stale response overwrite a newer one, and clearing an
  amount discards the pending request instead of letting it land on an empty form.
- Failed invoice/receipt PDF downloads show a visible, row-scoped error instead of being swallowed.

## Validation

- `npm run lint`, `CI=true npm test -- --runInBand` (58 suites, 567 tests), `npm run build:lib`,
  `npm run widget:dev`
- new regression tests were verified adversarially: the fix was reverted, the test confirmed to fail,
  and the fix restored
- reviewed in fourteen full passes against the complete diff rather than the latest change, because
  several fixes turned out to undercut earlier ones; the final pass reports no findings

## Two things a reviewer should know

**The visual baselines need regenerating.** The confirmation screenshot has no committed baseline
yet, and the payment-details capture was widened to include the recipient, so its existing image is
stale. Both need to be produced on a machine that can run the stack:

```
REACT_APP_API_URL=<api> npx playwright test e2e/buy-process.spec.ts \
  -g "should forward the personal IBAN selector and display Bank Frick details" \
  --project=chromium --update-snapshots
```

They were not generated here because the test drives a real token against a real API. The recipient
is additionally asserted as text, which holds regardless of the images. Tracked as #1195.

**Two behaviours are covered manually, not by tests.** The selector surviving a login redirect after
an expired token, and a logged-out customer's selector not reaching the next customer after Back plus
a reload. This repository mocks the SDK's authentication in its test setup, so no honest automated
test reaches that boundary; approximating it would have produced a test that only looks like proof.
Both are noted in the README and tracked as #1196.

## Dependency and rollout

Depends on DFXswiss/api#4384. Deploy the API first, then this consumer; an older API ignores the
request field, which this app now detects and surfaces instead of accepting silently.

dfxswiss/packages#191 is merged and published, and this PR consumes it: react 1.5.0 to 1.6.0, the
three local bridge types removed, `personalIbanProvider` typed by the SDK enum so the compiler
enforces the contract. The version jump also carries two unrelated SDK changes (#189, #193).


TaprootFreak:
Completed 6 full independent review passes. The final pass had 0 findings in both conformity and logic lanes. Latest GitHub CI is green, including lint, 39 test suites / 434 tests, app build, strict widget build, CodeQL, and the PR review bot.

TaprootFreak:
The earlier comment on this PR predates the current head and no longer describes it — its test counts are also outdated. Since then the diff went through 7 further full passes against the complete diff.

The substantial change is that the selector is no longer held in component state. Four different ways of storing it each had a path where state and source drifted apart, and on every one of those the customer requested a personal IBAN and silently received the default bank. It is now derived from its source on every read — the URL in the browser, the live property in the widget — which removes the class of bug rather than its instances.

The last pass found a contract mismatch worth noting: two error tokens had been renamed on the API side and were never followed here, so the checks never matched and the customer saw the raw token. A full comparison against the API enum found two further mismatches. Tests now pin every token, including negative cases for the retired spellings, so a rename on either side fails here instead of degrading silently.

Head fa2180a, all 6 checks green.

TaprootFreak:
Follow-up to the previous comment, after a factual correction from the product side: EUR accounts at Bank Frick stay with DFX; only the IBAN is the customer own. CHF accounts at Yapeal genuinely are customer-held.

The self-service screen promised the customer own name regardless of currency, and defaulted to EUR when none was given — reachable through the sitemap link, which passes none. That default was harmless until this feature, because there was no EUR provider. A missing currency is now a visible error rather than a silent default, and the confirmation text states what is true for the currency at hand. The promotional block on the buy screen already excluded EUR and is unchanged. Head 0d28571, all 6 checks green.

TaprootFreak:
Follow-up: the SDK release is out and this PR now consumes it rather than deferring it — react 1.5.0 to 1.6.0, and the three local bridge types removed. personalIbanProvider is the SDK enum now, so the contract with the API is compiler-enforced instead of duplicated by hand.

Worth flagging for review: the version jump also carries two unrelated SDK changes (#189, #193). They are not part of this feature but come with the version.

Verified locally with a clean install before pushing (lint, 53 suites / 515 tests, library build); the widget build is covered by CI. Head 218eb94, all 6 checks green.

TaprootFreak:
Fourteen full review passes against the complete diff against the merge base rather than the incremental change — several fixes in this series turned out to undercut earlier ones, and a delta review would not have caught that. The final pass reports no findings.

The two decisions that shaped this PR both came out of repeated failures rather than up front. The selector keeps no state at all, because every way of storing it had a path where the copy and its source drift apart and the customer silently receives the default bank. And the customer is now asked before a personal IBAN is requested, because the same URL can mean a leftover from the previous customer or a deliberate request by the current one — five attempts at detecting that from session state each closed one edge case and opened the next. A real, non-revocable account is created at the bank, so asking is both the honest and the proportionate answer.

Every response to an explicit request is verified against what was asked for; declining the confirmation fetches fresh selector-free details rather than displaying the response that was just rejected; and the feature-specific error wording applies only when the request actually carried the provider — it previously replaced the generic KYC dialog, with its working action, for ordinary bank customers.

One defect is worth naming because no command could have caught it: a commit captured the widget build's temporary entry-point swap, so the app would have compiled cleanly and served an empty page. Lint, 567 tests and both builds were green. It was found by reading the diff.

Every fix was verified adversarially: reverted, the test confirmed to fail, then restored.

Verified on the current head: lint, 59 suites / 569 tests, library build and widget build clean. CI green on this revision.

Two items are deliberately not in this PR and are tracked: #1195 (regenerate the two visual baselines) and #1196 (cover the two authentication-boundary behaviours end to end — this repository mocks the SDK's authentication, so an honest automated test cannot reach that boundary today).

TaprootFreak:
Two changes since the previous comment, both after the rebase onto the current develop.

The rebase itself brought five commits of other work under this branch. The conflicts were additive — the SDK version, the lockfile and translation entries — but the two things worth naming were not in the conflict markers: the branch now takes the newer `@dfx.swiss/react` from develop rather than pinning 1.6.0, and the full suite was re-run on the new base (61 suites, 612 tests).

Then the analysis bots: three `isRunning` conditions in the buy-info effect are removed. They sit in synchronous statements where the flag has just been set and cannot have changed; the asynchronous guards in `then`, `catch` and `finally` stay, because those genuinely observe a stale run.

The CodeQL alert on the selector read is a false positive and is marked as such at the line, with the reasoning: the public parameter carries a provider identifier — `frick` — never an IBAN, the value is normalised to the SDK enum, and neither the hook nor its context persists anything. Confirmation decisions live in an in-memory map keyed by customer id and are gone on remount. The alert itself is left for a maintainer to judge rather than dismissed.

CI green on this head.

PR #1191, opened 2026-07-25
feat(support): let customers type the receiver IBAN instead of picking one

TaprootFreak:
## Why

When a customer reports that a transfer never arrived, the support form asks which account it went to. For `TRANSACTION_MISSING` that field was a **required dropdown** fed by `GET /bank`, which lists only the collective accounts.

A customer who deposits through a **personal IBAN** could never find their real receiver there — and because the field is required, they had to pick an account that is not theirs. That wrong value lands in the ticket and reads to support like a statement from the customer. There are over 3,000 personal IBANs in production, so this is the normal case, not the edge case.

## What

The field is now free text, backed by `checkReceiveIban` from `useBank()`.

**The check is advisory and never blocks submission** — a support ticket must never fail on an unrecognized IBAN. The field stays required only in the sense of "must not be empty"; a blank-only entry is rejected as empty, which the dropdown made impossible before.

The hints, all phrased without blaming the customer:

| Status | Hint |
|---|---|
| recognized | We have recognized this IBAN. |
| not recognized | We could not recognize this IBAN with the information available to us. You can submit your request anyway. |
| not a valid IBAN | This does not look like a valid IBAN. |
| login needed | Please log in so that we can also check your personal IBAN. |
| any failure, incl. HTTP 429 | We could not check this IBAN at the moment. You can submit your request anyway. |

`NotMatched` claims no more than the status supports. Per the library's JSDoc it is not a statement that DFX does not own the IBAN: another customer's personal IBAN is never checked, and one left behind by an account merge is not either. So the text says we could not *recognize* it with the information available to us — not that it could not be assigned, and certainly not that the customer mistyped. Telling the very customer this change was built for to re-check an IBAN that is already correct would push them back to guessing, the exact failure this removes.

**Timing of the check, and of the hints.** The call is debounced by 500 ms and only fires once the value reaches 15 characters after normalization — the endpoint is rate-limited and the limit is shared by everyone behind the same network, and the API answers incomplete input with `InvalidIban` by design. Normalization strips every non-alphanumeric character, matching what the API does before it looks an IBAN up, so a grouped or dash-separated entry is measured and sent exactly as the server will read it. The debounce sits on that normalized value, so retyping a space costs no request.

Because 15 characters is the shortest IBAN in existence and a Swiss one needs 21, the threshold proves length but not completeness. Display is therefore gated on focus: while the field is focused only the positive confirmation appears — it can occur only for a complete IBAN that matched — and every other verdict waits until the field is left. That keeps a wrong "this does not look like a valid IBAN" off the screen mid-typing and stops the form from jumping. Submitting by mouse still shows the hint, because focus leaves the input on `mousedown`, before the click is dispatched; submitting with Enter from inside the field shows none, which is acceptable for a check that blocks nothing.

Stale responses are discarded through a request counter, and the check refuses to run at all when the debounced value is no longer the value in the field, so no hint can appear under a value it was not computed for.

`getBanks()` and its state and effect are gone from this screen — `banks` had no other use here, and the error banner that used to appear when the bank list failed to load goes with it. The sender IBAN, date and transaction fields are untouched.

## Dependency — the prerelease is the intended pin

**DFXswiss/api#4391 and DFXswiss/packages#192 are both merged.** The client call arrives here
through `@dfx.swiss/react`, pinned to `1.7.0-beta.0`; `@dfx.swiss/core` follows transitively to
`0.6.0-beta.0` and is not a direct dependency here.

The prerelease is the deliberate choice for this change, not a placeholder waiting on something
else: no release of the library to `main` is planned for it. Nothing about this PR is blocked on a
stable version, and reviewers should not hold it for one. Should a stable `1.7.0` be published
later, the range `^1.7.0-beta.0` accepts it, so an ordinary install picks it up without a second
range edit.

CI installs from the lockfile rather than the range, which is why the pin has to ship as part of
this change rather than only as a range bump.

## What was verified locally, and how

These were run against the published package, not a local build of it.

| Command | Result |
|---|---|
| `npm run lint` | exit 0 |
| `npm run build:dev` (the pipeline's real TS compile) | exit 0, build folder ready |
| `npm run widget:dev` (the stricter build) | exit 0, widget folder ready |
| `CI=true npm test -- --watchAll=false` | 39 suites, 470 tests passed |
| `prettier --check` on every changed and added file | clean |
| dependency pin | `@dfx.swiss/react` 1.7.0-beta.0, `@dfx.swiss/core` 0.6.0-beta.0 transitively |
| `eslint` on the added test files | 0 problems |

Note for reviewers: `npm run lint` globs `{src,apps,libs,test}/**/*.ts` and therefore does **not** cover `.tsx` — the only changed code file. It was linted explicitly and is clean, but a green lint job alone says nothing about this change.

## Tests

The check had no coverage at all. Two things are pinned, because everything else about
this feature is cosmetic next to them: **the check never blocks, delays or alters a
support request**, and **a hint never belongs to a value that has left the field**.

`src/__tests__/support-issue-receiver-iban.test.tsx` renders the screen and covers:

| Behaviour | What the test holds |
|---|---|
| Field visibility | free text appears only for `TRANSACTION_MISSING` |
| Threshold | 14 normalized characters make no request, 15 make exactly one |
| Normalization | separators do not count towards the threshold and are not sent; a grouped IBAN arrives as the server reads it |
| Debounce | fast typing produces exactly one request, for the final value |
| Formatting-only edit | regrouping an already checked IBAN costs no second request |
| Reset | any edit clears hint, spinner and unavailable state immediately, before a new answer arrives |
| Stale answer | a late response for an earlier value cannot overwrite the current hint — whether it resolves or rejects |
| Bound to its value | a verdict is shown only while the field still holds the IBAN it was computed for |
| Superseded value | no request is made for a debounced value the customer has already changed |
| Focus gating | positive confirmation shows at once; every other verdict waits for blur |
| Hint texts | the four texts and the colour that separates the positive one |
| Failure | a rejected check, rate limiting included, produces the unavailable hint and no form error |
| Spinner | on while in flight, off on success and on failure |
| Never blocks | the form is actually submitted while a check is still in flight, after a failure, for a non-IBAN entry, and for every one of the four verdicts |
| Never a form error | a rejected check produces no error banner and no field error, for rate limiting and for a server error alike |
| Raw value | the submitted request carries the customer's text unchanged, spaces and all |
| Required | an untouched or blank-only field is rejected as empty; any non-empty value stays submittable |
| Reason change | leaving `TRANSACTION_MISSING` clears the state and starts no further check |

Each of these was checked by breaking the screen on purpose and confirming the named
test turns red — the guards, the threshold, the gating, the normalization, the
required rule, the failure path and the submit handler were each removed in turn.

Two things resist a single-line mutation, and neither is a gap in the tests.
`ReceiverIbanCheckDelay` set to `0` changes nothing, because `useDebounce` computes
`delay || 500` and a falsy `0` becomes 500 again. And the reset effect shares its
three `setState` calls with the check effect's own guard branch, which runs on the
same dependency; removing a reset from one place leaves the other to compensate, so
only removing it from both turns the test red.

`src/__tests__/support-issue-translations.test.ts` pins the five new keys in de, fr and
it. It is deliberately scoped to those keys: the three files already differ on unrelated
entries, so a full parity test would fail on other people's gaps.

`@dfx.swiss/react` and `@dfx.swiss/react-components` are mocked rather than loaded —
their published ESM does not load under this repo's test command. The mocks reproduce
the rule merging and validation shapes the screen relies on and wire the input through
react-hook-form, so focus, validation and the watched value behave as in the app.

## Visual regression baselines

`e2e/support-issue-receiver-iban.spec.ts` captures the six visible states of the field:
empty, checking, recognized, not recognized, invalid, and check unavailable. It follows
the pattern of `e2e/realunit-support.spec.ts` — real customer auth, synthetic fixtures,
and `page.route` intercepting only `PUT /v1/bank/receiveIban` while everything else goes
to the API. Type and reason come from the screen's own `issue-type` and `reason` URL
parameters, so no dropdown interaction is needed.

**The baseline images are not in this commit.** `playwright.config.ts` puts the platform
in the snapshot path and all existing baselines are `*-chromium-darwin.png`;
`CONTRIBUTING.md` asks for new ones to be generated on the same platform as the existing
ones. Generating them here would have produced `*-chromium-linux.png` with different font
rendering — different files sitting next to the existing set, which is worse than none.

To add them, on macOS with the local API stack up and a `@dfx.swiss/react` that has
`checkReceiveIban` installed:

```
REACT_APP_API_URL=http://localhost:3000 npx playwright test support-issue-receiver-iban.spec.ts --project=chromium --update-snapshots
```

This writes all six baselines in one pass. Per `CONTRIBUTING.md` these tests are a
review aid and do not run in CI, so their absence blocks nothing.


TaprootFreak:
This diff went through **four review passes** before two independent reviews came back with no findings. What they caught:

**The length threshold did not do what it was added for.** Fifteen characters is the shortest IBAN in existence and a Swiss one needs twenty-one, so a customer pausing at sixteen still triggered a check and still read "this does not look like a valid IBAN" while typing a correct one — the very behaviour the threshold was meant to prevent. It moved the problem rather than solving it. A country-length table in the frontend was considered and rejected: it would put a second authority on IBAN validity next to the API. Instead, display is now gated on focus. Only the positive confirmation appears while the field is focused, because it can occur only for a complete IBAN that matched; every other verdict waits until the field is left.

**The threshold was measuring the wrong string.** It stripped only ASCII spaces while the API strips every non-alphanumeric character, so a dash-grouped entry counted its separators towards the length and was checked far too early. Normalization now matches the API exactly and is the single source for the threshold, the request and the equality guard.

**A check could start for a value no longer in the field.** If the debounce timer queued a value while the field changed, both updates could land in one commit and the response would write a hint under the new or empty value. Rather than reason about batching, the effect now refuses to run unless the debounced value is still the current one.

**The unmatched wording overstated twice.** It first asked the customer to check whether the IBAN was correct — which imputes a mistake in exactly the cases the API documents as not mistakes: another customer's personal IBAN is never checked, and neither is one left behind by an account merge. Telling the customer this change was built for to re-check a correct IBAN would push them back to guessing. The text now says only that we could not recognize it with the information available to us, and that the request can be submitted anyway.

Both reviews independently verified the assurance that matters most here: the check never enters form validation, never disables submission, and the ticket endpoint accepts any string. An unrecognized IBAN cannot stop a customer from reporting a problem.

`npm run lint`, `npm run build:dev`, `npm run widget:dev` and 427 tests pass locally against the pending library build.

**This PR stays in draft, and its CI will fail until the library release lands.** See the ordering section above — the lockfile has to be updated in this branch after the release, which is a required step and not an automatic one.

TaprootFreak:
Six review passes were needed before both independent reviews came back clean. The screen itself carried one real defect, and most of the rest were tests that did not hold what their names promised.

**The defect.** The check result was cleared in an effect on the normalized value, and effects run after the render. On the first render after an edit the previous verdict was therefore still in state and was shown under the new value — most visibly the positive confirmation, which is deliberately shown while the field still has focus. The result now carries the IBAN it was computed for, and a hint is derived only while the field still holds that value. Clearing it in the effect became unnecessary and is gone, together with the duplicate clearing in the check effect's guard branch.

**Mocks that were plausible but not faithful.** Three of them, each found by mutating the screen and watching the suite stay green:

- The `Form` mock attached `onSubmit` to the native form element, which the real component does not — it forwards `onSubmit` only to named children. Removing the submit button's own handler therefore passed.
- `Validations.Iban` was mocked as always passing, so adding real IBAN validation to the receiver field passed — while in the app it would reject a free-text entry and block a support request, which is the one thing this screen must never do.
- The submit button mock ignored `isLoading`, while the real component disables itself on `disabled || isLoading`. Tying the button's loading state to the check therefore passed.

**Guarantees that were pinned for one case only.** Submission was proven unaffected for an unrecognized IBAN, but not for the positive verdict, not for a non-IBAN entry, and not for a failure other than rate limiting. Each of those is now covered, so a verdict-specific or error-specific special case fails a test. The translation wiring is pinned too: removing the `translate` call around the hints used to leave the suite green, and every non-English customer would have seen English text.

Two reported points were rejected with reasons rather than applied. Retrying anonymously on HTTP 401 cannot occur here, because the endpoint's guard treats an unusable token as an anonymous request rather than rejecting it. And reproducing the button's `hidden` prop in the mock would not help: the real component adds a CSS class, which has no effect in jsdom, so the mock would have to use the HTML attribute instead — a less faithful mock that merely happens to be observable.

The PR description was brought to the end state along the way: the ordering note now reflects that the API and library changes have landed, and the dependency pin is described for what it is.

TaprootFreak:
The seven missing baselines for `e2e/support-issue-receiver-iban.spec.ts` are in — plus one fix that running the spec for the first time turned up.

**The spec could not work as committed.** `/support/issue` sits behind `useKycLevelGuard`, and a freshly registered account is below that level, so the app redirects to contact data and the Receiver IBAN field never renders. All seven tests timed out waiting for a field that was not there. The route setup now pins `kyc.level` in the `GET /v2/user` response, next to the `receiveIban` mock the tests already install. Nothing changes on the server, so the account the other specs share keeps the state they rely on, and the screenshots do not depend on whatever level the local account happens to have. The seven tests themselves are untouched — same selectors, same mock states, same waits, same screenshot names and options.

**How the images were produced and checked**

- Generated on macOS against a local API stack, per CONTRIBUTING → *Visual regression tests*; the files carry the `*-chromium-darwin.png` suffix like the existing ones, and no other baseline was regenerated.
- `--update-snapshots`: 7/7 green.
- A second run **without** `--update-snapshots`: 7/7 green against exactly those files, with identical sha256 before and after — a run that is only green while writing proves nothing.
- The local account was at KYC level 0 both before and after the run, which is the evidence that this leaves no server-side state behind.
- Every image was opened and checked against the state its test claims: **01** empty field with placeholder, no hint; **02** filled and focused, the in-field spinner present and still no hint (in a still image the spinner is a small mark, since the comparison disables animations); **03** the green *We have recognized this IBAN.* while the field still has focus; **04–07** the grey hint for not matched, invalid, check unavailable and login required, each after blur.
- No real data: local seed data, the ISO 13616 example IBAN, placeholder name and date.

**One thing worth knowing.** The *Date of the transaction* field renders `new Date()` as its placeholder, so all seven images contain the date they were taken on. Regenerating them on another day produces a small diff there. I measured it rather than guessed: that text is 174 pixels of ink, so even a completely different date changes about 350 pixels against the `maxDiffPixels: 5000` each test allows — roughly a factor of 14 of headroom. Pinning the browser clock would remove it, but that would touch all seven tests for a difference below the tolerance, and the cause is in the screen itself rather than in this test. Happy to do it if you would rather have byte-stable baselines.

Playwright does not run in CI by design (CONTRIBUTING), so these files cannot affect the checks; CI is green.

PR #1192, opened 2026-07-27
test(safe): document the Safe screen with baseline screenshots

TaprootFreak:
## Why

The Safe screen (`/safe`) had no Playwright spec and no baselines at all. The only coverage was
a smoke test in `payment-and-special.spec.ts` that asserts the `body` is visible. A UI change to
this feature therefore produced no visual diff for a reviewer to look at.

Per `CONTRIBUTING.md` these tests run locally only and are deliberately **not** wired into CI —
they are a review aid, not a regression gate.

## Coverage

Eight distinct states, each a real state change:

| Screenshot | State |
| --- | --- |
| `01-safe-portfolio` | loaded portfolio, asset list, transaction interface, activity list |
| `02-safe-portfolio-chf` | currency switch to CHF |
| `03-safe-portfolio-usd` | currency switch to USD |
| `04-safe-deposit-crypto` | deposit, type crypto (receive form) |
| `05-safe-withdraw-fiat` | withdraw, type fiat |
| `06-safe-withdraw-crypto` | withdraw, type crypto (send form) |
| `07-safe-swap` | swap interface |
| `08-safe-pdf-modal` | PDF statement date selection |

Two states are intentionally not captured separately: "Deposit / Fiat" is the initial state of
the transaction interface and is visible in 01-03, and the activity list sits below the
interface and appears in every shot. Capturing them again produced byte-identical images.

## Two things worth knowing about the setup

**Viewport height.** The screen scrolls inside an `overflow-auto` container, so `fullPage: true`
still yields a 1280x720 image. Without a taller viewport the deposit, withdraw and swap
screenshots contain only the portfolio and the top edge of the form — the very thing they are
supposed to document. `responsive.spec.ts` already sets viewports per test, so this follows
existing practice.

**Screenshot tolerance.** `maxDiffPixels` is kept small on purpose. With a large budget,
`--update-snapshots` silently keeps an outdated baseline when the content has changed; the
images then document a state that no longer exists. That was observed while preparing this
change, which is why the budget covers anti-aliasing only.

## Verification

Generated on macOS (`*-chromium-darwin.png`) against a local API per the CONTRIBUTING quick
start, with a funded custody test portfolio and a set of custody orders so the portfolio,
asset list and activity list are all populated.

- All eight images have distinct checksums, i.e. no screenshot is a duplicate of another.
- A second run without `--update-snapshots` passes, so the baselines are reproducible at the
  chosen tolerance.

## Known limit of this change

The operation screenshots capture the **forms**, not the payment information that the backend
generates once a form is submitted (receive QR/address, send quote, bank details). Capturing
those would prove the custody quote endpoints still work, which these screenshots do not.

Two states are simply not reachable locally at all: creating a swap is rejected by the API with
`"This transaction can only be created manually by support"`. The others are reachable in
principle — a receive order created directly against the local API returns a complete
`paymentInfo` — but driving them reliably through the UI needs selectors for the generated
views that are worth getting right rather than guessing at. That is a sensible follow-up on top
of this baseline set, not a reason to hold back the coverage that exists.

Everything committed here is verified: the suite passes, a second run without
`--update-snapshots` passes as well, all eight images have distinct checksums, and the clock is
pinned so the PDF date cannot drift.


TaprootFreak:
Three review passes against the complete diff. Each one found a real defect:

**Pass 1** — the form screenshots showed only the portfolio. The screen scrolls inside an
`overflow-auto` container, so `fullPage: true` still produced 1280x720 images; two baselines were
byte-identical. Fixed with a taller viewport, following the per-test viewports in
`responsive.spec.ts`.

**Pass 2** — two findings. The PDF modal prefills the current date, so the committed baseline
carried a moving value and would have rotted; the clock is now pinned before navigation. The
second finding is described below and is deliberately not fixed here.

**Pass 3** — verification of the above.

Verified for the final state: the suite passes, a second run without `--update-snapshots` also
passes, and all eight baselines have distinct checksums.

### One review point left open on purpose

The operation screenshots capture the forms, not the payment information the backend generates
after submission (receive QR/address, send quote, bank details). Capturing those would also prove
the custody quote endpoints still work, which these screenshots do not.

Creating a swap is rejected outright by the API ("this transaction can only be created manually by
support"), so that state is not reachable locally at all. The others are reachable in principle —
a receive order created directly against the local API returns a complete payment info — but
driving them through the UI needs selectors for the generated views that are worth getting right
rather than guessing at. That is a follow-up on top of this baseline set; it does not reduce the
value of the coverage added here, which replaces having no baselines at all for this screen.

PR #1193, opened 2026-07-27
feat(safe): open the Safe without a preselected transaction area

TaprootFreak:
Opening the Safe put the deposit form on screen before the user had asked for anything: `SafeTransactionInterface` initialised `mode` to `TransactionMode.DEPOSIT`, so the Fiat/Crypto switch, the amount field and the NEXT button were all rendered on arrival.

Now **Deposit**, **Withdraw** and **Swap** render as plain text with nothing selected, and the area below them stays empty until one is chosen. After that everything behaves as before, including the Fiat/Crypto switch for Deposit and Withdraw.

## Two details worth flagging

**The unselected style would have been wrong here.** In header mode, entries that are not selected get `text-dfxBlue-500/30` — 30 % opacity. With nothing selected, all three would have looked disabled rather than available, so `ButtonGroup` renders them at full strength in that state. Once a selection exists the existing look is untouched: chosen one highlighted, the others dimmed.

**`selected` now accepts `undefined`.** `ButtonGroup` is also used for the currency switch, the chart timeframe, the Fiat/Crypto switch and a metric switch on another screen; all four pass a real value and are unaffected — widening the type does not change their behaviour, and the new branch only applies when nothing is selected.

Two files, +6/−3. Lint clean, no type errors in application code.


TaprootFreak:
Ready for review after three review passes to zero findings.

## What the passes found

**1. The Safe baselines were stale.** Screenshots 01–03 documented the screen with the deposit form open, which is no longer the default. They are regenerated from the same synthetic dataset, so the only difference in the image is the now empty transaction area. 04–08 are untouched, because those steps click into a mode first — which is also the check that the change does nothing beyond the default state.

**2. Deposit/Fiat lost its coverage.** It used to be the preselected state and was therefore documented by the portfolio shots. With nothing preselected it would have been the only interface without a baseline, so a regression in it could pass review unseen. It now has its own shot, placed right after the portfolio states — where a user meets it on the first click.

Step 1 additionally asserts that neither the `Type:` row nor the `Next` button exists before a mode is chosen. Without that, a regression could put the form back in front of the user and only the image would notice.

## One thing worth recording

The baselines are generated from a local dataset, not from anything in this repository. Regenerating them with whatever happens to be in the local database is therefore not safe — it can pull unrelated or sensitive data into a public repository. The dataset was restored to the one the existing baselines were made from before regenerating, which is why the diff is limited to the intended change.

## Verified

The full spec passes against the committed baselines in a clean run, without `--update-snapshots`.

PR #1194, opened 2026-07-28
feat(safe): let a user switch between the Safe accounts they may see

TaprootFreak:
## What this changes

The Safe screen only ever showed the signed-in user's own holdings. Someone granted access to another person's Safe could not reach it — the data is there and the API serves it, but no screen offered a way to ask for it.

An **account picker** appears whenever more than one account is reachable. The whole screen follows the chosen account: portfolio value, chart, holdings, recent activity and the PDF statement. With a single reachable account the picker is not rendered and the screen behaves exactly as before, which is what almost every user has today.

On an account held by **inspection only**, the deposit / withdraw / swap section is not rendered at all — not greyed out, not disabled. The API refuses those operations anyway (DFXswiss/api#4417); offering buttons that fail on click would be worse than not offering them.

## How it works

- `GET /custody/account` returns own and shared accounts with the caller's level per account. The picker pre-selects the account the caller owns; with read-only access only, the first.
- When an account is selected, balance, history, orders and statement come from `/custody/account/:id/…`.
- **The legacy Safe appears in that list with a `null` id.** It has no account row and is read through the plain endpoints (`custody`, `custody/history`, `custody/order`, `custody/pdf`). One helper decides the path, so this cannot be overlooked per call site.
- The selection is held as a stable key rather than the account object — an object in a dependency list is a new reference on every render.
- Each load is tied to the account it was started for and discards its result if the selection has moved on, so a late answer cannot paint one account's holdings under another account's name. The manual reload after a completed order shares that guard.
- The statement download refuses while the account list is still loading instead of quietly serving the caller's own Safe, and the transaction area stays hidden until the access level is known.

## Verification

- `tsc --noEmit`, `prettier --check`, `npm run lint` and `CI=true npm test` (427 tests) all clean.
- Exercised against a running instance in three shapes:
  - **legacy Safe** — the picker is absent, the plain endpoints are used, no failed requests, the transaction area is offered;
  - **one owned plus two shared read-only accounts** — switching reloads the view, shared accounts are marked, and the transaction section is absent on both;
  - **rapid switching back and forth** — the view ends on the account actually selected, with its own figures.
- Against the committed `safe.spec.ts` baselines the differing pixels are 3434 of a 1280×1800 image, all of them numbers from a local dataset, with every element at its original position — the single-account and legacy layouts are unchanged, so those baselines are not regenerated.

## New baselines

`e2e/safe-accounts.spec.ts` covers the three states that only exist with more than one account. The third asserts `toHaveCount(0)` for all three transaction buttons before taking the picture — a screenshot alone would not prove absence. The dataset behind those images is invented; this repository is public.

## Relation to #772

#772 targeted the same feature and has been closed. It called `safe-account/*`, a path that has never existed — the endpoints are `custody/account/*` — and it changed only `safe.dto.ts` and `safe.hook.ts`, without the components that would render a picker. Its groundwork for creating accounts and listing grants is noted there for whoever builds the grant-management screens, which are still missing.


TaprootFreak:
Seven review passes so far, thirteen findings, all fixed. Three of them would have reached customers:

**The Safe was broken for every legacy user.** The account list returns the legacy Safe as an entry with a null id — I had assumed the list would simply be empty. The screen therefore requested `custody/account/null/…` and showed nothing but an error. Legacy is nearly every account today. My local dataset always had account rows, so I had only ever exercised the new case, never the old one.

**A mandate over someone else's account offered the full transaction area** — but the order endpoints carry no account, so the order would have been booked against the caller's own holdings while the screen named another account. Acting now requires one's own Safe; acting on someone else's behalf does not exist in the backend at all.

**Switching accounts mid-request could paint one account's holdings under another's name.** Guarding by "is this account still selected" was not enough: switching away and back leaves the same selection in place, so a slow first answer still won. Each kind of load now carries a generation and only the newest may commit.

The rest: a wrapper that stayed in the DOM and shifted every single-account layout by 40px, an endless request loop, a manual reload path that skipped the guards, and smaller matters — stale comments, an unused import, a rule-less eslint-disable, a one-way mount flag.

Beyond CI, each fix was verified against a running instance rather than only in tests: the legacy Safe loading through the plain endpoints, rapid A→B→A landing on the right account with its own figures, and a write grant on a foreign account offering no transaction area while the caller's own still does.

PR #1197, opened 2026-07-28
fix(safe): label a foreign write mandate as view only, and guard the rule in CI

TaprootFreak:
Follow-up to #1194, from findings in its final review pass.

## The inconsistency

The account picker labelled entries by access level alone:

```ts
descriptionFunc={(account) => (account.accessLevel === 'Read' ? 'View only' : '')}
```

A **write grant on someone else's account** therefore appeared exactly like one's own — no marking, apparently fully usable. Selecting it made the deposit / withdraw / swap section disappear without a word, because `canTransact` also requires ownership.

That state is reachable: `grantAccess` accepts `Write` for any user on any account. Label and behaviour now share one predicate, `canActOn`, so the list cannot promise what the screen refuses.

## The missing guard

`canActOn` is the only thing standing between a mandate holder and an order booked against the wrong Safe — orders carry no account, so acting while looking at someone else's Safe books against the caller's own. Acting on another person's behalf does not exist in the backend at all; `initiatedBy` is a column nothing writes.

That rule went wrong twice during the review of #1194, and its only coverage was a Playwright spec — which is explicitly **not** a CI gate. It has a unit test now, five cases:

| account | level | may act |
| --- | --- | --- |
| own | write | yes |
| own, narrowed by the owner | read | no |
| foreign | read | no |
| **foreign** | **write** | **no** |
| legacy | write | yes |

## Why the predicate moved out of the hook

It is plain logic with no React in it, and importing the hook pulls in ESM that Jest will not parse — the test suite failed to run at all. `src/util/safe-account.ts` makes it testable and puts it where the rest of the repo keeps such helpers.

## A note on how ownership is determined

`isOwnAccount` still infers ownership from the absence of the `owner` field. That is a side effect of how the API loads its relations, not a contract — DFXswiss/api#4424 adds an explicit `isOwner` flag, and this should switch to it once that lands. The comment says so.

## Verification

```
CI=true npm test        → 38 suites, 432 tests
npm run lint            → clean
tsc --noEmit            → clean
```

The `safe-accounts` spec now carries a third account in its dataset — shared, held under a **write** mandate — and a fourth screenshot of exactly that state: labelled "View only", holdings visible, no transaction area, despite the grant saying write. That is the state this pull request is about, and it had no picture before.

Two existing baselines changed in the same run. `02` gained the third row now that the picker lists three accounts — expected. `03` differs only in the order two assets appear in, which has nothing to do with this change: the balances are loaded without an ORDER BY, so their order is whatever the database returns. DFXswiss/api#4428 fixes that; until it lands, that shot is not stable across runs.


TaprootFreak:
Four review passes to zero findings.

The substantive one was the missing picture. This pull request is about a single state — a write mandate over someone else's Safe — and no baseline showed it; the only evidence was a sentence saying it had been checked by hand. A third account now carries that mandate and the fourth screenshot is that state.

Regenerating those shots turned up something that looked like noise and was not: one baseline differed only in the order two positions appeared in, at identical values. The balances are loaded without an ORDER BY, so their order is whatever the database returns — the same holdings can reshuffle between two requests for no reason a customer could perceive. DFXswiss/api#4428 fixes that.

Also worth recording: the pull request text claimed the baselines were unchanged long after they had changed, because an earlier correction silently failed to apply and I reported it as done without checking. The text now matches the images, verified against what the API actually returns.

PR #1206, opened 2026-07-29
Issue the personal IBAN directly from the selector

TaprootFreak:
## Summary

A customer who opens `/buy?asset-in=EUR&personal-iban=frick` gets the personal IBAN
without being asked again. Setting the selector *is* the request; the intermediate
prompt interrupted the purchase without adding a decision.

Removed:

- the confirmation prompt component and its copy
- the per-customer confirmed/declined state that survived for the app instance
- the widget occurrence mechanism, whose only purpose was to force a fresh
  confirmation when the selector changed

Kept, deliberately:

- the acknowledgement shown when the selector cannot be honoured — ineligible
  currency or payment method, or a response that failed verification. Without it a
  customer who asked for a personal IBAN would silently receive shared-account
  details instead.
- response verification and unrecognized-selector handling

## What now stands between the selector and an account

Issuance creates a real account at Bank Frick that cannot be undone. Two conditions
remain: an initialised wallet and an authenticated customer. Nothing else.

Both are a single `||` of negations, so either one alone must block the request. No
test pinned them independently before — an implementation checking only one would
have passed. Two tests now cover each half, verified by flipping the operator and
confirming exactly those tests fail.

## Consequence worth stating plainly

With the prompt gone there is no step at which the customer sees what is about to
happen. The buy screen fills in currency, payment method, asset and a default amount
by itself, so for an authenticated customer the account is created about half a
second after the page loads, without a single click. On the payment-info screen there
is no debounce at all.

That also means the selector need not come from the customer: a link carrying it,
opened by someone already signed in, issues the account. This is the accepted
trade-off for removing the prompt, not an oversight — recorded here so it is a
decision on the record rather than a surprise later.

## Validation

- `CI=true npm test -- --runInBand` — 58 suites, 599 tests
- `npm run lint`, `npm run build:lib` — clean
- test accounting, measured on both sides rather than estimated: the merge base runs 62 suites
  / 617 tests, this branch 58 / 599 — 4 suites and 18 tests fewer. The drop is the confirmation
  and occurrence coverage that no longer has a mechanism to cover; what replaced it is the
  selector-hook and gating coverage described above.
- reviewed against the full diff, not the increment


TaprootFreak:
Six review passes were needed to reach zero findings.

Rounds one to three settled the substance: the end-to-end spec still drove the deleted prompt, the real selector hook had lost all direct coverage, and the replacement test turned out not to discriminate a realistic break — it stayed green when the authenticated-customer check was replaced with a bare "is there a session" check. That was only established by applying the break and watching the suite, which is now the standard the added fixtures are held to.

Round four surfaced the one real defect: the check read the account field out of the decoded session, which survives token expiry, so an expired session reported an authenticated customer. Not exploitable — the endpoint that issues the IBAN rejects an expired token before anything is created — but with the confirmation step gone this check carries more weight, so it now consults the freshness flag as well.

Rounds five and six found only statements claiming more than the code delivers: two in the README about what the tests cover, a test named after an expired token that passes the flag by hand, a comment describing a value as set when it is recomputed, and the test counts in this description, which were wrong three times before being measured on both sides.

PR #1210, opened 2026-07-29
a687c0ab - fix(safe): format activity timestamps in Swiss notation

TaprootFreak:
## Problem

The Safe activity list rendered order timestamps in US notation:

```
Deposit
Completed · 1/28/26, 1:31 PM
```

`formatTimestamp()` passed the locale from `useSettingsContext()` into `toLocaleString()`, so the date format followed the app UI language — with an English interface that resolves to `en-US`, giving month/day/year and a 12-hour clock.

The date format is a Swiss convention. It is not a translation of the interface and must not follow the selected language.

## Change

The locale is fixed to `de-CH`, and `formatTimestamp()` no longer takes a `locale` parameter:

```
Deposit
Completed · 28.01.2026, 13:31
```

Explicit field options are used instead of `dateStyle: 'short'`, which would shorten the year to two digits (`28.01.26`).

The timestamp selection itself is unchanged: a completed order shows `completedAt ?? created`, every other state shows `created`, and a row with neither shows no date rather than `Invalid Date`.

## Baselines

Twelve screenshots are regenerated: the eight of `safe.spec.ts` and the four of `safe-accounts.spec.ts`. Both specs render the same `/safe` route and therefore the same activity list. The PDF modal shot of `safe.spec.ts` is unchanged — the modal covers the activity list.

All were regenerated on the same seed data the previous baselines were taken on, so the diff isolates the date format: every amount, asset and total is identical to before.

## Notes

- `npm run lint` does not cover this change — its glob is `{src,apps,libs,test}/**/*.ts` and therefore matches no `.tsx` file, i.e. no React component in the repository. The change is covered by `build:dev` and `widget:dev`, both of which type-check it. There is no test file for this component and none was added.
- **Safe baselines are no longer deterministic while an sZCHF position is in the seed data.** The saving position accrues interest server-side against `new Date()`, which the spec's `page.clock.setFixedTime()` does not affect — two runs minutes apart produce different values. These baselines were taken with the interest rate set to zero locally so the diff stays isolated. Anyone regenerating a Safe baseline in the future will hit this; it deserves its own fix and is out of scope here.
- Other `toLocaleString()` call sites without an explicit locale exist elsewhere in the app (compliance helpers, RealUnit screens, account screen). They have the same issue but are deliberately out of scope.


TaprootFreak:
Three review passes were needed before both dimensions came back clean.

- **Pass 1** flagged that the committed Playwright baselines still showed the old US format. They were regenerated locally against the API on the same seed data, so the diff isolates the date format.
- **Pass 2** found four more affected baselines: `safe-accounts.spec.ts` renders the same `/safe` route and therefore the same activity list, but never names the component — only the route. Those four were regenerated as well, bringing the total to twelve.
- **Pass 3** came back with no findings, including an explicit check that no further spec renders a timestamp that would need regenerating.

One observation worth recording for later, unrelated to this change: the Safe baselines are no longer reproducible while an sZCHF position sits in the seed data. That position accrues interest server-side against `new Date()`, which the spec's frozen clock does not affect — two runs minutes apart produce different values. These baselines were taken with the interest rate set to zero locally. It deserves its own fix.

Issue #1208, opened 2026-07-29
e2e: wallet-personal-iban-session.test.tsx is flaky and fails PRs at random


TaprootFreak:
`src/__tests__/wallet-personal-iban-session.test.tsx` fails non-deterministically on `develop` itself, with no change applied. It turns "Build and test" red on unrelated pull requests, where it reads as a regression caused by the PR.

## Evidence

Five consecutive runs on plain `develop` (3aa225d), macOS, node 20.20.2, `CI=true npm test -- --watchAll=false --testPathPattern=wallet-personal-iban-session`:

| Run | Result |
| --- | --- |
| 1 | 1 failed, 2 passed |
| 2 | 2 failed, 1 passed |
| 3 | 3 passed |
| 4 | 1 failed, 2 passed |
| 5 | 3 passed |

Three of five runs red, on an unmodified checkout. The same suite also failed once in GitHub Actions and passed on a plain re-run of the identical commit.

## The failing assertion

Both affected cases are in `widget session initialization with a personal-IBAN selector`:

- `logs out a persisted customer for a malformed non-JWT session and remains usable`
- `logs out a persisted customer for a JWT without a customer identity and remains usable`

```
expect(jest.fn()).not.toHaveBeenCalled()

Expected number of calls: 0
Received number of calls: 1

1: {"customer": 103}

  216 |       expect(screen.getByTestId('decision')).toHaveTextContent('required'),
  217 |     );
> 218 |     expect(mockReceiveQuote).not.toHaveBeenCalled();
```

The preceding `waitFor` only asserts that the decision element reached `required`. It does not establish that the session teardown finished, so the quote call from the stale customer can still land between the two statements. The assertion is racing the very effect it means to exclude.

## Why it matters

The suite is part of the "Build and test" gate. A red run here costs every author a diagnosis pass to establish that their change was not the cause, and it trains reviewers to wave red checks through — which is worse than the flake itself.

TaprootFreak:
Gone from `develop`, but by removal rather than repair — worth recording which, because the two are not the same for anyone who hits this next.

**The file no longer exists.** #1206 (merged today, `8155f5a5`) deleted `src/__tests__/wallet-personal-iban-session.test.tsx` along with the confirmation mechanism it covered — the pre-confirmation prompt, the per-customer confirmed/declined state and the widget occurrence mechanism all went with it. The reproduction above ran on `3aa225d2`, before that merge.

[...]

PR #1213, opened 2026-07-29
a0b32b4e - Give async waits the time a loaded machine needs

### LINKED ISSUE #1212: test: a 1s default wait makes findBy* a coin flip under full parallel load
`src/__tests__/transaction-document-error.test.tsx` fails non-deterministically in a full suite run on `develop` itself, with no change applied. It turns "Build and test" red on unrelated pull requests, where it reads as a regression caused by the PR.

## Evidence

Five consecutive full-suite runs on plain `develop` (932a288e), macOS, node 20.20.2, `CI=true npm test -- --watchAll=false`:

| Run | Result |
| --- | --- |
| 1 | 604 passed |
| 2 | 604 passed |
| 3 | **1 failed**, 603 passed |
| 4 | 604 passed |
| 5 | 604 passed |

The failing case is `TransactionList document error display › maps StoredBankNoLongerAcceptsPayments on Open receipt to customer-facing copy`:

```
Unable to find role="button" and name "Open receipt"

  203 |     renderList();
  204 |
> 205 |     const openReceipt = await screen.findByRole('button', { name: 'Open receipt' });
```

At that moment the DOM holds the list heading and nothing else — the transactions have not rendered yet.

## Root cause

Not a race in the test. The same file passed **20 of 20 isolated runs** (`--testPathPattern=transaction-document-error`). `TransactionList` has no timer in its load path and both mocks resolve immediately.

Jest runs the suite across one worker per core, so a single worker can go without CPU for well over a second. Instrumenting every `findBy*` in the file and raising DOM Testing Library's `asyncUtilTimeout` to 15000ms, across ten full runs on an 18-core machine:

| Metric | Value |
| --- | --- |
| Measurements | 50 |
| Median wait | 278 ms |
| Longest wait | 1523 ms |
| Over the 1000 ms default | 7 of 50 (14%) |

All ten runs passed (6040 tests, 0 failures) with the wait limit as the only changed variable. So the limit is what decides the outcome, and the 1000 ms default that DOM Testing Library applies is simply below what a loaded machine delivers.

## Why it matters

This is not confined to one file — it applies to every `findBy*` and `waitFor` in the suite. `transaction-document-error.test.tsx` (added in #1190) is only where it surfaced first. A red run costs every author a diagnosis pass to establish that their change was not the cause, and it trains reviewers to wave red checks through — which is worse than the flake itself.

Same failure class as #1208, different file: there the test raced its own teardown, here it races the scheduler.

TaprootFreak:
Closes #1212.

`transaction-document-error.test.tsx` failed 1 in 5 full-suite runs on plain `develop`, with nothing changed: `findByRole('button', { name: 'Open receipt' })` ran into DOM Testing Library's 1000 ms default while the list had not rendered yet. A red "Build and test" that a plain re-run clears costs every author a diagnosis pass, and teaches reviewers to wave red checks through.

## The limit is the cause, not the test

- **20 of 20 isolated runs** of the file pass. `TransactionList` has no timer in its load path and both mocks resolve immediately — there is no race to fix in the test.
- Instrumenting every `findBy*` in the file across **ten full suite runs** on an 18-core machine:

  | Metric | Value |
  | --- | --- |
  | Measurements | 50 |
  | Median wait | 278 ms |
  | Longest wait | 1523 ms |
  | Over the 1000 ms default | 7 of 50 (14 %) |

- With the wait limit as the **only** changed variable, all ten runs passed: 6040 tests, 0 failures.

Jest runs one worker fewer than the machine has cores, and under that parallelism a single worker goes without CPU for well over a second at a time. The default assumes it does not.

## What changes

`src/setupTests.ts` only:

- `asyncUtilTimeout` 1000 → 5000 ms. The longest wait measured was 1523 ms, so the new limit sits well clear of everything observed. The margin comes from the measurements rather than from a guess.
- `jest.setTimeout` 5000 → 15000 ms. It has to stay above `asyncUtilTimeout`, or Jest ends the test first and Testing Library's DOM dump naming the missing element is replaced by a bare "Exceeded timeout of 5000 ms". `react-scripts` does not accept `testTimeout` in the `jest` section of `package.json`, so it is set here.

`tsconfig.build.json` gains one exclude entry. It kept test code out by name pattern alone — `**/*.test.ts`, `**/*spec.ts`, `**/__tests__/**` — and `src/setupTests.ts` matches none of them, so it stayed in the program that `build:lib` emits to `dist/` and that the review bot type-checks. That was invisible while the file held nothing but an import: `jest.setTimeout` is the first value access to `jest` in it, and `tsconfig.json` sets `typeRoots: ["typings"]`, which keeps `node_modules/@types` out of scope, so `@types/jest` never loads and the reference fails with TS2708. The whole test suite hits the same wall, which is why it is excluded from this config to begin with — this file was simply missed.

No assertion is weakened. Each one still waits for exactly the condition it waited for before, and a genuinely broken expectation still fails — it just gets the scheduler time it needs first. The setting is suite-wide on purpose: every `findBy*` and `waitFor` in all 604 tests runs against the same limit, and `transaction-document-error.test.tsx` (added in #1190) is only where it surfaced first.

## Verification

Ten full suite runs on this branch, same machine and command as the measurements above. Plus the four gates that matter here, including the one that caught the TS2708 above: `tsc -p tsconfig.build.json --noEmit` (no errors, one before), `npm run lint`, `npm run build:dev`, and the suite at 604 of 604.

Related: #1208 covered a different flake in the same class — a test racing its own teardown rather than the scheduler. That file was removed with #1206; the issue is closed with the evidence.

### DIFF
--- a/src/setupTests.ts
+++ b/src/setupTests.ts
@@ -3,3 +3,19 @@
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom';
+import { configure } from '@testing-library/react';
+
+// Jest defaults to maxWorkers = CPU cores - 1 (17 workers on 18 cores); under that load a
+// worker can go without CPU time for over a second at times.
+// Measured findBy* waits across 10 full suite runs (50 samples): median 278ms, maximum 1523ms;
+// 7 of 50 exceeded 1000ms (DOM Testing Library's default async timeout). This does not weaken
+// any assertion: every assertion still checks exactly the same condition, and a genuinely
+// broken expectation still fails — this setting only gives the scheduler the time it needs.
+// Applies to the whole suite, not only the file where the flakiness was first noticed.
+configure({ asyncUtilTimeout: 5000 });
+
+// Must stay above asyncUtilTimeout so Jest does not abort first and replace Testing Library's
+// informative DOM dump (e.g. "Unable to find role=button …") with a bare "Exceeded timeout of
+// 5000 ms" and no diagnostic detail. react-scripts does not accept testTimeout in the jest
+// section of package.json (supportedKeys in createJestConfig.js omits it), so set it here.
+jest.setTimeout(15000);
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -1,4 +1,4 @@
 {
   "extends": "./tsconfig.json",
-  "exclude": ["node_modules", "test", "dist", "**/*spec.ts", "**/*.test.ts", "**/__tests__/**"]
+  "exclude": ["node_modules", "test", "dist", "**/*spec.ts", "**/*.test.ts", "**/__tests__/**", "src/setupTests.ts"]
 }


TaprootFreak:
Three review passes across two lenses, conformance and logic. The code was confirmed unchanged in all of them — both findings landed in the description rather than the diff:

1. "Three times the longest wait measured" — 5000 ms is 3.28 times 1523 ms, not three times.
2. The replacement wording ("3.3 times it, rounded to a whole number") contradicted itself, since 3.3 is not a whole number.

The description now gives the measured maximum and the new limit without a ratio, which is what the numbers actually support.

The logic pass took the central question head-on — whether a longer wait can turn a test green that ought to fail — and answered it from the tree rather than from principle: `waitForElementToBeRemoved` appears nowhere in the repository, exactly one negative assertion sits inside a `waitFor` callback (`realunit-compliance-screen.test.tsx:150`, and it checks a converged end state), and no `waitFor`/`findBy*` passes its own `timeout:` argument that would bypass the global setting.

One point is left standing deliberately: the commit message still says "three times the longest wait measured" and carries the same imprecision. Fixing it would mean amending a commit that is already pushed, which this workflow does not permit. Squash merges here take the body from the commit messages, so whoever merges may want to reword that clause — the accurate version is that 5000 ms sits well clear of the 1523 ms longest wait measured.

TaprootFreak:
Correction to the comment above — that state was not ready to review. The bot had already flagged `src/setupTests.ts(21,1): error TS2708: Cannot use namespace 'jest' as a value` and I missed it: the bot posts a comment but does not fail its own check, so every entry in `gh pr checks` read SUCCESS, and neither `npm run lint` nor `build:dev` covers this file.

The error was real. `tsconfig.build.json` keeps test code out by name pattern alone — `**/*spec.ts`, `**/*.test.ts`, `**/__tests__/**` — and `src/setupTests.ts` matches none of them, so it stayed in the program that `build:lib` emits to `dist/` and that the bot type-checks. `tsconfig.json` sets `typeRoots: ["typings"]`, which keeps `node_modules/@types`, and with it `@types/jest`, out of scope. That stayed invisible while the file held nothing but an import; `jest.setTimeout` is the first value access to `jest` in it.

Excluding the file is the narrower fix, rather than widening `typeRoots`: this package publishes its types (`types: dist/public-api.d.ts`, `files: ["dist"]`), so pulling all of `node_modules/@types` into the library build risks leaking ambient globals — Jest's among them — into what consumers see. `tsc -p tsconfig.build.json --noEmit` now reports no errors where it reported one, confirmed on a clean `npm ci` and by the bot's own run on `1ba0b52f`.

Six review passes in total, four on conformance and two on logic. What they found: two wordings in the description claiming a precision the numbers did not support, and — after the bot surfaced it — the TypeScript error above. The logic pass settled the question that matters for a raised timeout structurally rather than by sampling: `waitFor` only succeeds by its callback not throwing, never by time running out, so a longer limit cannot turn a failing assertion green. It can only delay a correct failure.

One point stands unfixed deliberately: the first commit message says "three times the longest wait measured", where 5000 ms is 3.28 times 1523 ms. Amending a pushed commit is not something this workflow permits. Squash merges here take the body from the commit messages, so that clause is worth rewording at merge time.

PR #1221, opened 2026-07-30
baa7748e - Split the financial log entry type by response mode

TaprootFreak:
Frontend counterpart to DFXswiss/api#4519, which is why that PR exists at all: the Financial
Overview screen is open around the clock and reloads its three-day window every minute.

## What changed on the API side

`GET /dashboard/financial/log?byType=false` now returns only `timestamp`, `totalBalanceChf` and
`btcPriceChf` per entry. The remaining fields are omitted in that mode — deriving them means
parsing a ~43 kB JSON document per data point, and the chart does not use them. The full mode
(`byType` omitted or `true`) is unchanged.

## Why this PR

Nothing is broken at runtime today. The overview screen hands its entries to
`TotalBalanceLongChart`, which reads exactly the three fields the chart mode returns; the
`plusBalanceChf`/`minusBalanceChf` values on that screen come from `getLatestBalance()`, a
different endpoint.

The problem is the type. `FinancialLogEntry` declared the omitted fields as guaranteed `number`,
so it promised something the chart mode does not deliver. Reaching for `entry.plusBalanceChf` on
the overview screen would compile cleanly and yield `undefined`.

## The change

The type is split rather than widened:

- `FinancialLogChartEntry` — the three fields the chart mode returns.
- `FinancialLogEntry extends FinancialLogChartEntry` — adds the full-mode fields.

Making the fields optional instead would have pushed the problem into the history screen, which
legitimately relies on them being present.

The mode is now chosen by which hook function you call — `getFinancialLogChart()` versus
`getFinancialLog()` — instead of by a boolean argument whose value the return type ignored. A
caller can no longer request one shape and be typed as the other.

The history screen and its components (`total-balance-short-chart`, `balance-by-type-area-chart`,
`financial-log-table`) keep using `FinancialLogEntry` and are untouched, because the split is
additive to the existing full-mode type.

## Verification

Type errors in `.tsx` files surface in the build rather than in lint, so `npm run build:dev` is
the gate that matters here. Result will be posted below.

### DIFF
--- a/src/components/dashboard/total-balance-long-chart.tsx
+++ b/src/components/dashboard/total-balance-long-chart.tsx
@@ -1,11 +1,11 @@
 import { ApexOptions } from 'apexcharts';
 import { useMemo } from 'react';
 import Chart from 'react-apexcharts';
-import { FinancialLogEntry } from 'src/dto/dashboard.dto';
+import { FinancialLogChartEntry } from 'src/dto/dashboard.dto';
 import { TimeRange } from 'src/util/chart';
 
 interface TotalBalanceLongChartProps {
-  entries: FinancialLogEntry[];
+  entries: FinancialLogChartEntry[];
   timeRange?: TimeRange;
   dark?: boolean;
 }
--- a/src/dto/dashboard.dto.ts
+++ b/src/dto/dashboard.dto.ts
@@ -1,9 +1,14 @@
-export interface FinancialLogEntry {
+/** Chart mode: GET /dashboard/financial/log?byType=false — only these three fields are present. */
+export interface FinancialLogChartEntry {
   timestamp: string;
   totalBalanceChf: number;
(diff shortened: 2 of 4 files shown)


TaprootFreak:
Verified against the full pipeline gates: lint and `build:dev` both pass. The build is what matters
here, since type errors in `.tsx` files surface there rather than in lint.

Scope check on the two callers of the endpoint: the overview screen reads only the three fields the
chart mode returns, and the history screen uses the full mode unchanged. So this changes no runtime
behaviour — it makes the types say what the API actually delivers.

One note for reviewers: `total-balance-long-chart.tsx` is flagged by Prettier, but it already is on
`develop`. Reformatting it here would have buried a four-line type change in an unrelated diff.

The review pass over this branch is still running. Anything it turns up will be handled in a
follow-up PR rather than held against this one.

PR #1222, opened 2026-07-30
fix(format): render dates in a fixed Swiss notation, not the browser locale

### LINKED ISSUE #1214: Date formatting still follows the UI language outside the Safe
## Problem

#1210 fixed the Safe activity list: its timestamps are now formatted with a fixed `de-CH` locale, because the date format is a Swiss convention and must not follow the selected interface language.

The rest of the app still does follow it. These call sites pass no locale at all, so `toLocaleString()` falls back to the browser locale — an English UI renders `6/12/26, 2:30 PM` instead of `12.06.2026, 14:30`:

- `src/util/compliance-helpers.tsx:74, 137, 194, 207, 211`
- `src/screens/account.screen.tsx:273`
- `src/screens/realunit.screen.tsx:160, 255, 320`
- `src/screens/realunit-quotes.screen.tsx:73`
- `src/screens/realunit-quote-detail.screen.tsx:119`
- `src/screens/realunit-transaction-detail.screen.tsx:111`

These are compliance and RealUnit views, i.e. internal tooling rather than the customer-facing product — which is why they were deliberately left out of #1210 rather than fixed silently.

## Suggested fix

Same approach as `src/components/safe/transaction-history.tsx`: fixed `de-CH` with explicit field options.

```ts
toLocaleString('de-CH', { day: '2-digit', month: '2-digit', year: 'numeric', hour: '2-digit', minute: '2-digit' })
```

Explicit options rather than `dateStyle: 'short'`, which shortens the year to two digits (`28.01.26`).

A shared helper would be better than repeating the option object twelve times — `src/util/utils.ts` already hosts `formatCurrency` and would be the natural home. `src/components/dashboard/financial-log-table.tsx:11-19` has a comparable local formatter that could fold into it.

## Out of scope

The PDF dialog in the Safe uses a native `<input type="date">`, whose display format is controlled by the browser and system locale and cannot be set through `Intl` options. Changing that would mean replacing the native control with a custom one — a separate decision.

Follow-up to #1210.

Danswar:
Closes #1214. Follow-up to #1210, which fixed the Safe activity list.

## Problem

Date formatting is a Swiss convention, not a translation of the interface. Call sites across the app passed no locale to `toLocaleString()`, so the format followed the browser locale — an English UI rendered `6/12/2026, 2:30:45 PM` instead of `12.06.2026, 14:30:45`.

## Change

Four helpers in `src/util/utils.ts`, next to the existing `formatChf()`/`formatCurrency()`:

```ts
formatSwissDate(value)                 // 12.06.2026
formatSwissTime(value)                 // 14:30
formatSwissDateTime(value)             // 12.06.2026, 14:30
formatSwissDateTimeWithSeconds(value)  // 12.06.2026, 14:30:45
```

Explicit field options rather than `dateStyle: 'short'`, which would shorten the year to two digits.

The compliance `formatDate()`/`formatDateTime()` helpers delegate to them. That is where most of the leverage is: they are imported across 24 files, so their callers are fixed without touching each import.

**Four helpers, not one, because precision is preserved rather than normalised.** A bare `toLocaleString()` emits seconds. Every site that showed seconds still shows them; every site that had explicitly chosen minute precision keeps it. Normalising everything to the `{day, month, year, hour, minute}` set the issue suggests would have silently dropped seconds from ~28 compliance call sites — note lists, call-queue comments, issue tables — where events can share a minute and the timestamp is what orders them. The Safe therefore still renders `12.06.2026, 14:30`, exactly as #1210 shipped it.

The rule across this change: **notation is fixed, precision is untouched, timezone is untouched.** No helper sets `timeZone`, so every site still renders in the viewer's local zone.

## Scope

The issue listed 12 call sites. The same defect class turned out to be wider:

- 8 further locale-less sites the issue's list missed — `realunit-transactions:75`, `realunit-user:115,190`, `compliance-kyc-step:114,180`, `compliance-recommendation-graph:350`, `transaction.screen:1083,1219`
- 8 compliance sites that passed `de-CH` but no field options, rendering `12.6.2026` next to the padded `12.06.2026` — among them a KYC step's `created`, rendered `12.6.2026` on the review panels and in the browser's own format on the Stammdaten panel
- `dashboard-financial-live:58`, a timestamp card passing `de-CH` without field options
- `support-stats.ts:104`, whose day label followed the browser, and `recommendations-section.tsx:209`, whose expiry date followed the UI language
- `support-tickets.screen.tsx:112` and `chat.screen.tsx:485`
- 5 compliance screens carrying their own `formatDate()` with an option set identical to the shared helpers; they delegate now, and the three that guard missing values keep their `'-'` guards
- `safe/transaction-history.tsx`, which held the inline original of this option set since #1210. It delegates now; the option set is byte-identical, so the Safe renders exactly as before.

Two of these are worth calling out.

**The transaction detail is converted because the account list links into it.** Converting only the list would have left the same timestamp reading two ways one navigation apart, where before this change both sides followed the browser and at least agreed.

**The chat message time asked for `hour12: false` against the browser locale.** That is not a 24-hour clock: on locales defaulting to `h24` it renders midnight as `24:30`, so an `en-US` browser disagreed with every converted site for an hour a day — one click from the converted ticket list. It uses `formatSwissTime()` now.

### One deliberate exception: month names

`trendLabel()` in `support-stats.ts` took no locale on its month branch — the branch the default 365-day period actually renders. A month name is a word, not notation, so it follows the interface language (as the app's other month labels do) while the numeric day label is pinned to Swiss. `locale` is an explicit parameter of `trendLabel()`.

`computeStatistics()` stores raw stable bucket keys and the screen localizes them in a render-time memo. Loading them per language instead would make a language switch a refetch: the chart would blank behind a spinner, and a refetch that failed would replace statistics that were already correct with an error page.

## Left alone

**`financial-log-table.tsx` keeps its local formatter.** It sets `timeZone: 'UTC'` and its header reads "Timestamp (UTC)". A shared helper without that option would not change the formatting — it would change the value by two hours in CEST.

**`formatDateTimeShort()` keeps its own option set.** The two-digit year is the point of that variant; it only gains the fixed locale.

**`transaction.screen.tsx:1017` is not converted.** `step.timestamp` is typed `Date` but is a raw ISO string at runtime — the API layer parses JSON without a reviver — so the call lands on `String.prototype.toLocaleString()` and returns the string unchanged, in UTC and identical in every locale. It was never part of this defect class, and converting it would move the instant into the viewer's zone and drop both the `Z` marker and the milliseconds. An earlier commit converted it; it is reverted, with a comment.

**Still out of scope:** `payment-link.screen.tsx:517`, `payment-link-pos.screen.tsx:261-262`, `payment-routes.screen.tsx:648,1451` and `chat.screen.tsx:235`. No converted screen links to the payment screens directly; they are reachable only through the admin sitemap, which the global nav offers on every screen. Converting them is not purely mechanical either — `payment-link-pos.screen.tsx:261` renders a bare `toLocaleTimeString()`, so preserving its precision would need a seconds-bearing time helper. They are better done together, separately.

`chat.screen.tsx:235` renders weekday and month *names*, which by this change's own rule should follow the interface language, and it already does.

## Visual baselines — NOT regenerated

Six specs render a changed line — 14 shots: `compliance.spec.ts` (8), `compliance-recommendation-graph.spec.ts` (1), `realunit-compliance.spec.ts` (1), `realunit-support.spec.ts` (2), `support-dashboard-overview.spec.ts` (1) and `support-dashboard.spec.ts` (1).

Carve-outs, so the counts can be checked against the 11 + 4 + 2 + 2 + 2 + 1 shots those specs actually take:

- the recommendation-graph spec's other three shots switch the detail panel to nodes with no `tradeApprovalDate`, so that row stays `-`
- `realunit-compliance`'s search shot renders no date at all
- `compliance.spec.ts`'s two search shots render collapsible sections that default to closed, and its KYC-stats shot renders numbers only
- `support-dashboard-overview`'s statistics shot does render a changed code path (`trendLabel`'s month branch), but at the default 365-day period it reproduces its old output: the app locale is derived from the browser language, and both resolve the same short month name on any machine that took the existing baseline
- `user-flows.spec.ts`'s account shot leaves the converted row collapsed

Converting the transaction detail additionally affects `refund-flow.spec.ts` and `buy-crypto-fail-refund.spec.ts` (8 baselines each). Only their date-bearing shots move; I have not enumerated which, since that needs the specs actually run.

**Update:** the local stack was stood up (`docker compose up -d` + `npm run setup` in a sibling `api` checkout) and each of the six specs was actually run.

Three read `formatDate`/`formatDateTime`/`trendLabel` output through a real backend and need named fixtures — a specific customer, tickets in specific states — that no available seed script produces: `compliance.spec.ts`, `support-dashboard-overview.spec.ts`, `support-dashboard.spec.ts`. Manufacturing those by hand (registering a user to land a specific ID, driving tickets into specific states) felt like the wrong thing to improvise inside this PR, so they stay stale.

The other three mock their feature data via `page.route()` and only needed a working admin login. Of their 11 shots, 10 stayed within this repo's own 5000px diff tolerance — the date-string length change is real but small — so nothing to regenerate there per CONTRIBUTING.md's "only regenerate what your change actually affects." The eleventh, `realunit-compliance`'s dossier shot, does cross it, but for an unrelated reason: independent of this change, that baseline is already stale against the current app (it's missing a "Checks" column and other fields the screen has gained since), so regenerating it now would bury this change inside a confusing, unrelated diff. Left alone.

`realunit-support-01-list.png` is the one shot that's both affected and cleanly regenerable — verified pixel-for-pixel that only the Created/Last Msg columns changed, everything else identical — and is regenerated and committed.

The two refund specs (`refund-flow.spec.ts`, `buy-crypto-fail-refund.spec.ts`) remain unregenerated: they need `TEST_TX_ID`/`BUY_CRYPTO_TX_ID` pointing at transactions in specific failure states that no seed script creates either.

**A real, pre-existing gap surfaced along the way:** `e2e/test-wallet.ts`'s hardcoded sign-message constant doesn't include the `[loc]_` environment prefix the API now requires (`Config.auth.signMessagePrefix`, non-empty for every environment except PRD) — so the documented local workflow (`REACT_APP_API_URL=http://localhost:3000 npx playwright test ...`) fails admin auth outright for anyone following CONTRIBUTING.md fresh, on any spec, unrelated to this PR. Worked around locally to get the one baseline above; not fixed here, since it's shared test infrastructure and out of scope for a date-formatting change. Worth its own PR.

The Safe baselines from #1210 are unaffected, since the Safe output is unchanged.

## Corrections to intermediate commit messages

Commits are not amended after push, so claims the later rounds found false are corrected here, and this description is the accurate one. The code in each commit is unaffected — the errors are in the justifications.

- `c172264` says every call site outside the Safe passed no locale. Many passed `de-CH` without field options; `9f388a6` and `d9bd6f6` fixed those. Its worked example, `6/12/26, 2:30 PM`, is the `dateStyle: 'short'` output; a bare `toLocaleString()` renders `6/12/2026, 2:30:45 PM` — the seconds being exactly what `9f388a6` then had to restore.
- `9f388a6` says two sites "followed the UI language". One followed the browser; only the recommendation expiry followed the UI language.
- `d9bd6f6` justifies the live-dashboard change with a log table below the card. No such table is rendered anywhere — the component has no importers. It also says all five delegating screens keep their `'-'` guards; only three ever had one.
- `65bd773` says nothing this change converts navigates into the out-of-scope screens; the converted ticket list does link into the chat. It also says three transaction detail sites are converted and all four preserve their field set — `7b77d25` reverted one, so two are converted, and the ticket list is date-only with no seconds to preserve.
- `7b77d25` says the other two sites on the transaction screen are "real Date values". They are typed `Date` and are ISO strings at runtime for the same reason it gives for the reverted one; they format correctly because both wrap in `new Date(...)` first.

## Verification

`npm run lint`, `npm test` (612 passed, 60 suites), `npm run build:dev`, `npm run widget:dev` — all green locally on Node 20.

Unit tests cover all four helpers and both `trendLabel` branches, including the midnight case that motivated the chat change. They are built from local date components so they hold in any timezone; the suite was additionally run under `TZ=Pacific/Kiritimati` and `TZ=America/Buenos_Aires`.

Prettier (3.9.6) reports the same nine files at this HEAD as at `origin/develop` across this diff's 34 files — identical set on both sides, so the pre-existing drift is untouched and no new drift is introduced. Prettier is not a CI gate in this repo.


Danswar:
8 review passes to 0 defects across both lanes (conformity/CONTRIBUTING.md and logic/correctness), run in parallel with CI.

Summary of what those rounds caught, beyond the initial conversion:
- a site that looked locale-dependent but wasn't (`transaction.screen.tsx:1017` — typed `Date`, raw ISO string at runtime; converting it would have moved a UTC instant into the viewer's timezone)
- a request race and a response-shape guard in the support statistics view, introduced and then fixed across the later rounds
- a chat timestamp that wasn't actually pinned to 24h (`hour12: false` renders `24:30` at midnight on some locales)
- several corrections to earlier commit messages' stated reasoning, recorded in the PR description since commits aren't amended after push

Known gap, stated in the description: ~16 visual baselines are stale and not regenerated here — the specs need `ADMIN_SEED`/`TEST_TX_ID` fixtures not available in this environment.

PR #1227, opened 2026-07-31
Report the errors customers are shown

TaprootFreak:
## Problem

When the app shows *"Oh, sorry, something went wrong"*, nothing is recorded. `error.screen.tsx` is the router's `errorElement`, which means it is the one place where the error that broke the render is still in hand — and it read none of it. `useRouteError` was never called anywhere in the app, the error object was discarded, and the failure existed nowhere but in the customer's browser console.

The practical consequence: this screen can only ever be known through the support channel, never measured, and a fix for it cannot be verified.

A second defect sits underneath. The chunk-reload guard in `src/index.tsx` listened on `window` `error` and `unhandledrejection`. A failed `React.lazy` import reaches neither: React catches it and hands it to the error boundary. **So the recovery never ran for the case it was written for** — and every screen except the home screen is lazy-loaded, which is why the failure shows up on navigation rather than on first load.

## Change

- `ErrorScreen` reads `useRouteError()` and reports the error once, with type, message, stack, route and build id. The route comes from the router, not from `window.location`: the widget and library builds run on a memory router, where the browser URL belongs to the host page and never changes as the customer moves between screens.
- The chunk-reload guard moves to that same boundary, where the error actually arrives.
- Detection stays on the bundler's own wordings, and that is sufficient: a stale chunk answered with the app shell makes the script load without registering the chunk, and the bundler's loader turns that into a `ChunkLoadError` reading `Loading chunk N failed. (missing: <url>)`. Verified against the loader in the built bundle.
- It briefly also matched the bare syntax errors a browser produces while parsing that HTML as a script — but one of those, `Unexpected token '<'`, is exactly what `JSON.parse` reports when a response is HTML instead of JSON. A gateway, WAF or login redirect answering an API call would then have reloaded the page and discarded whatever the customer had typed. A regression test holds that line.
- The `window` listeners stay for failures outside the router (startup, an `import()` in a click handler). They are installed from `index.tsx`, the standalone app's entry point, and deliberately not from the two embedded ones — see below.
- Reporting is fire-and-forget with `keepalive`, so a failing report never surfaces as a second error and survives the reload that may follow it. Fields are truncated to the limits the endpoint accepts, and an empty message gets a substitute — the endpoint requires one, and that rejection was being swallowed silently, which is the very blind spot this closes.
- Deliberate navigations to `/error?msg=...` are reported with their message. That path has no route, so the router previously produced a bare 404 that said nothing about the actual failure.

## Embedded, the host's page is not ours to reload

The widget and the library run inside someone else's page, and `window` belongs to that page — the closed shadow root isolates DOM and CSS, not `window`. Reloading it to recover a stale chunk would discard the host's forms, cart and scroll position over a deploy of ours.

Both embedded builds mark themselves, and there:

- **no reload** — the customer is told to reload the page instead, in all three languages, and the support button is left out of that case: the support screen is lazy-loaded like every other, so it would fail on the same chunk and send them in circles. Every other failure keeps the button, because a reload would not fix those.
- **no window listeners** — they are page-wide, so on a host that ships its own bundler they would catch that bundler's chunk failures and file them as ours. Recovery is off there anyway, and the failures worth seeing arrive through the router regardless.

## What `public/_redirects` actually does

The comment there claimed real files under `static/*` are matched before the SPA fallback. That holds only while they **exist**. A build asset that does not falls through:

```
curl -sI https://app.dfx.swiss/static/js/999.deadbeef.chunk.js
-> 200, content-type: text/html, cache-control: public, max-age=31536000, immutable
```

So a client asking for a chunk that a later deploy replaced does not get a 404 it could recover from — it gets the app shell parsed as JavaScript, cached hard. The comment now records that measurement; the app-side handling above recovers the client by reloading.

Closing it at the source needs a host-side change, since `_redirects` cannot express a 404 status and `_headers` cannot tell an existing asset apart from the fallback — both match on path alone. That is deliberately **not** part of this PR: it changes how every asset request is served and needs verification on a real deployment.

## Tests

44 new tests. The central one renders a route that throws inside a real router and asserts the error reaches the reporting call — the wiring that was missing. Also covered: route-not-found, the explicit-message path, report-once, the route reported under a memory router, chunk detection including the JSON-parsed-as-HTML regression, truncation, the reload guard including blocked storage and the embedded case, the embedded error screen, and the listener wiring itself for both channels.

Full suite: 62 suites, 648 tests, all green. App build and widget build both compile.

## Depends on

`DFXswiss/api#4550`, which adds the endpoint. It should land first; until then, reports fail silently by design.

### DIFF
--- a/public/_redirects
+++ b/public/_redirects
@@ -1,3 +1,17 @@
 # SPA deep-link fallback: unknown paths serve the app shell (client-side routing).
-# Real files (static/*, widget/*, favicon, ...) are matched by the host first.
+#
+# Real files under static/*, widget/*, favicon, ... are matched by the host before this rule --
+# but only while they EXIST. A build asset that does not exist falls through to this rule and is
+# answered with index.html and status 200, and the /static/* rule in _headers then stamps that
+# HTML with a year of immutable cache. Verify with a request for any absent asset:
+#   curl -sI https://app.dfx.swiss/static/js/999.deadbeef.chunk.js
+#   -> 200, content-type: text/html, cache-control: public, max-age=31536000, immutable
+#
+# This is what a client sees when it requests a chunk that a later deploy replaced: not a 404 it
+# could recover from, but an app shell parsed as JavaScript. The app treats that as a chunk load
+# failure and reloads once (src/util/client-error.ts), which recovers the client as long as the
+# bad response was not cached under a chunk URL that a later build reuses unchanged.
+#
+# Closing that gap needs a host-side change, since _redirects cannot express a status of 404 and
+# _headers cannot tell an existing asset apart from this fallback -- both match on path alone.
 /*    /index.html    200
--- a/src/Main.lib.tsx
+++ b/src/Main.lib.tsx
@@ -1,5 +1,12 @@
 import { createMemoryRouter } from 'react-router-dom';
 import App, { WidgetParams } from './App';
+import { installChunkErrorHandling, markEmbedded } from './util/client-error';
+
+// The library is imported straight into someone else's React app, so it has no entry point of its
+// own to wire this up — it has to do it here. Reporting only: a chunk failure is not worth
+// reloading the consumer's page over.
+markEmbedded();
+installChunkErrorHandling();
 
 function MainLib(params: WidgetParams) {
   return (
--- a/src/Main.lib.tsx
+++ b/src/Main.lib.tsx
@@ -1,12 +1,11 @@
 import { createMemoryRouter } from 'react-router-dom';
 import App, { WidgetParams } from './App';
-import { installChunkErrorHandling, markEmbedded } from './util/client-error';
+import { markEmbedded } from './util/client-error';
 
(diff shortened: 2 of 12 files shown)


TaprootFreak:
Six review passes to zero findings.

Substantive changes along the way: chunk detection briefly matched `Unexpected token '<'`, which is also what `JSON.parse` reports when a response is HTML instead of JSON — a gateway or login redirect would then have reloaded the page and discarded whatever the customer had typed. The reported route moved from `window.location` to the router, because the widget and library builds run on a memory router where the browser URL belongs to the host page. Recovery by reload was turned off for those builds entirely, since reloading a third party's page over a deploy of ours discards their state — and because that left the customer without a way out, the embedded error screen now asks them to reload instead of offering a support button that is itself lazy-loaded and would fail on the same chunk. The page-wide listeners are no longer installed there either; they would have caught a host bundler's chunk failures and filed them as ours.

CI green. 44 new tests, full suite 648.

PR #1230, opened 2026-08-01
Answer a missing build asset with 404, not the app shell

TaprootFreak:
## Problem

A request for a build asset that no longer exists is answered with the app shell and status 200:

```
curl -sI https://app.dfx.swiss/static/js/999.deadbeef.chunk.js
-> 200, content-type: text/html, cache-control: public, max-age=31536000, immutable
```

The client asked for JavaScript and received an HTML page — then cached it under the chunk's own URL
for a year, because the `/static/*` rule in `_headers` matches on path alone and cannot tell an
asset apart from the fallback.

This is what a customer hits after a deploy replaces the content-hashed chunks while their tab is
still open. #1227 made the app recover from it and made the failure visible; this closes it at the
source.

The widget is affected the same way, and worse in kind — its chunks load inside third-party pages:

```
curl -sI https://app.dfx.swiss/widget/v1.0-chunks/999.deadbeef.chunk.js
-> 200, content-type: text/html
```

## Change

A Pages Function on the asset paths asks for the asset and inspects what comes back. Nothing served
under those paths is HTML, so the content type is what separates a real asset from the fallback —
both are successful responses. A successful HTML response means the asset is gone, and the request
is answered `404` with `no-store`.

`public/_routes.json` restricts the Function to `/static/*` and `/widget/*`. Every other request —
the app shell, deep links, the root — is served statically and never reaches it.

**Conditional requests get the same treatment.** A client that stored the app shell under an asset
URL before this existed would otherwise have had that entry confirmed on revalidation: the binding
answers `304` against the fallback's own validator, and a `304` carries neither body nor content
type. On `304` the lookup is repeated without the conditional headers, so the path can be judged by
what it actually holds. A real asset keeps its `304` and the client keeps its valid copy; the
fallback becomes `404`; a failed retry passes through as the failure it is. The extra lookup happens
on revalidation only — a client whose copy is still fresh sends no request at all.

That last part matters beyond tidiness: it is the one case the app-side recovery in #1227 could not
reach.

## Measured, not assumed

Against a real `npm run build` output under `wrangler pages dev`:

| Request | Before | After |
|---|---|---|
| `main.<hash>.js`, a `.chunk.js`, a `.chunk.css` | 200 js/css | 200 js/css |
| **absent `/static/*` asset** | **200 text/html** | **404, no-store** |
| existing widget chunk | 200 js | 200 js |
| **absent widget chunk** | **200 text/html** | **404, no-store** |
| widget entry point | 200 | 200 |
| `/buy`, `/`, favicon, manifest | 200 | 200 |
| conditional GET, existing asset | 304 | 304 |
| **conditional GET, absent asset** | **304** | **404** |

Two approaches were tried first and rejected on evidence:

- **A 404 in `_redirects`.** Rejected by Cloudflare as an invalid rule; the absent asset still came
  back as HTML with 200.
- **Narrowing the fallback with an explicit `/static/*` rule.** This broke the site in the emulator:
  an explicit rule takes precedence over real files, so *existing* assets were served as HTML too.
  It would have been a full outage on deploy.

## Checks

The Function is deployed but was outside everything CI runs — a syntax error in it would have
reached production with a green build. ESLint now covers `functions/**/*.js`; the type-aware parser
cannot resolve files outside `tsconfig.json`, so that path is linted without type information, which
still catches what matters. Verified by introducing a syntax error and confirming the lint run fails
on it.

Nine tests pin every branch: the fallback becoming 404, a real asset passing through with its
headers, a missing content type, a non-HTML error status, a successful non-200 fallback, the
conditional path in all three of its outcomes, and an HTML error page keeping its status. Full
suite green. The Function stays out of the app bundle — the built output contains no reference to
it.

## What this costs

Cloudflare bills requests that invoke a Function; static asset requests are free. `_routes.json`
keeps that scope as small as the fix allows — the app shell and all navigation stay static — but
asset requests that were previously free now count. There is no way to have Pages return a 404 for a
missing asset without a Function.

## Before main

The emulator is not the edge. **This should go to DEV first** — `dfx-app-dev.pages.dev` reproduces
the defect identically today, so the same `curl` calls there confirm or refute the fix on real
infrastructure. Worth checking explicitly, because it is the one way this change could do real harm:
that an *existing* asset still returns `200 application/javascript` and not the app shell.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,17 @@ module.exports = {
     },
     // Story files are excluded from tsconfig.json, so the type-aware parser cannot resolve them.
     ignorePatterns: ['.eslintrc.js', '**/*.stories.tsx'],
+    overrides: [
+      {
+        // Cloudflare Pages Functions are deployed but never bundled into the app, so they sit
+        // outside tsconfig.json and the type-aware parser cannot resolve them. Linting them
+        // without type information still catches the failure that matters here: a syntax error in
+        // a file that reaches production without any other check ever reading it.
+        files: ['functions/**/*.js'],
+        parserOptions: { project: null },
+        env: { browser: true, node: false, jest: false },
+      },
+    ],
     rules: {
       '@typescript-eslint/interface-name-prefix': 'off',
       '@typescript-eslint/explicit-function-return-type': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,8 +20,8 @@ module.exports = {
       {
         // Cloudflare Pages Functions are deployed but never bundled into the app, so they sit
         // outside tsconfig.json and the type-aware parser cannot resolve them. Linting them
-        // without type information still catches the failure that matters here: a syntax error in
-        // a file that reaches production without any other check ever reading it.
+        // without type information still gives this path a static syntax check, which the build
+        // does not provide for files it never compiles.
         files: ['functions/**/*.js'],
         parserOptions: { project: null },
         env: { browser: true, node: false, jest: false },
--- a/functions/[[path]].js
+++ b/functions/[[path]].js
@@ -12,19 +12,46 @@
 export async function onRequest(context) {
   const response = await context.env.ASSETS.fetch(context.request);
 
+  // A conditional request can be answered 304 against the fallback's own validator, which would
+  // leave a client that stored the app shell under an asset URL holding on to it. A 304 carries
+  // neither body nor content type, so the only way to see what the path holds is to ask again
+  // without the conditions. This costs a second lookup on revalidation only -- a client whose copy
+  // is still fresh sends no request at all, and never reaches this code.
+  if (response.status === 304) {
+    return await resolveConditional(context);
+  }
+
   // Restricted to successful responses. An HTML error page is a failure with its own status, and
   // reporting it as 404 would state that the asset is permanently gone on the strength of what may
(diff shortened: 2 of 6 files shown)


TaprootFreak:
Six review passes were needed before this came back clean.

Rounds 1–3 were about accuracy of the comments: a claim that contradicted `_redirects`, a stated reload-versus-retry behaviour that does not exist in this codebase, a stale claim that nothing else reads the function, and an over-broad summary in the test file. All were cut back to what is verifiable rather than reworded.

Rounds 4–5 found two real defects in the logic. The check keyed on status 200 exactly, so a successful non-200 response carrying the fallback would have passed through as a valid asset. And conditional requests were not handled at all: a client that had already cached the app shell under an asset URL would have had that entry confirmed on revalidation — the very case this change exists to end. Handling it also closes the one gap the app-side recovery in #1227 could not reach. The fix for that then had a defect of its own, turning a failed retry into a 304, which round 5 caught.

One round-4 finding was rejected rather than fixed: that the catch-all rewrite shadows existing assets and would break the site. Measurement against a real build contradicts it — an existing asset returns 200 application/javascript with the function active, and production serves those assets as JavaScript today. It was not reported again once the measurements were put to it.

Along the way the function was brought under lint and test; it was deployed but outside every check CI runs.

Full suite: 63 suites, 665 tests, green. Please deploy to DEV before main — the same two curl calls there settle on real infrastructure what the emulator can only indicate.

PR #1232, opened 2026-08-01
Report a failure once, and repair the stored answer that caused it

TaprootFreak:
## What is happening in production

Since the reporting from #1227 went live, the ingest endpoint takes **~68 requests per second**,
around the clock. 99.9% of them are rejected by its rate limit — of 5'000 sampled requests, 4'998
came back `429`. The rest of the API sees ~1'100 requests per 10 minutes, so this one endpoint
carries roughly 40× the entire real traffic of the platform.

Reproduced in a real browser: **a single open page produces 19–30 reports per second**, indefinitely.
The customer sees one broken screen. We receive ~68'000 reports an hour from them.

## Defect 1 — the guard could not hold

`ErrorScreen` guarded against repeat reports with a `useRef`, which lives and dies with the mount.
The router remounts the same route repeatedly — measured as 357 × `history.replaceState('/buy')`
during 12 seconds of the loop — and every remount builds a fresh component with a fresh ref. React
then re-throws the cached chunk failure, the boundary mounts again, and the report goes out again.

The guard now lives in the module, keyed on type + message + route for one minute. It survives the
remount, because that is the thing it has to survive.

This matters beyond the noise: while the loop fills the rate limit, it **pushes out the reports of
every other customer**. The flood blinds exactly what it floods.

## Defect 2 — reload repeated the failure instead of ending it

A reload re-reads `index.html`, but not the stored answer for the chunk itself — and a chunk URL is
stable for as long as its content is. `2688.57cddf4b.chunk.js` is byte-identical across builds and
therefore keeps its name. A customer who once stored an app shell under that URL is served it again
on every single attempt, for the year `_headers` marks those assets fresh.

That is why the same customers kept failing on a file that demonstrably exists:

```
curl -sI https://app.dfx.swiss/static/js/2688.57cddf4b.chunk.js  ->  200, application/javascript
```

The failing address is now re-requested with `cache: 'reload'` before the reload. That forces the
browser to the network and makes intermediate caches revalidate, replacing the stored copy. The
reload then happens either way — it is the recovery, and it must not be skipped because the repair
failed.

Only same-origin addresses are touched. The address comes out of an error message, and an embedded
or third-party bundler could put someone else's host there.

## While we are here: `(error:` versus `(missing:`

Worth recording, because it sent this investigation down the wrong path once. Cloudflare sets
`x-content-type-options: nosniff` on every response. When HTML arrives under a `.js` URL the browser
therefore **refuses to execute it** and the script tag fires `error`, not `load` — which webpack
turns into `(error: <url>)`. Verified three ways: in the webpack source
(`JsonpChunkLoadingRuntimeModule.js:188`, `event.type === 'load' ? 'missing' : event.type`), in the
built production bundle, and by reproducing it in a browser.

So `(error:` does **not** rule out "HTML instead of JavaScript" — with `nosniff` it is that case's
signature. `(missing:` would only appear without `nosniff`.

## Tests

672 tests pass, 63 suites. New cases cover: the repair request and its exact options, the reload
still happening when the repair fails, a foreign-origin address being ignored, and all four
behaviours of the repeat guard (same failure suppressed, different route reported, different failure
reported, reported again after the window).

## What this does not fix

Entries already held in the **CDN edge cache**. A hit there never reaches the origin — measured on a
poisoned URL: `cf-cache-status: HIT`, `age: 22609`, a year of freshness. Those entries have to be
purged out of band, otherwise new visitors can still pick one up.

### DIFF
--- a/src/__tests__/client-error.test.ts
+++ b/src/__tests__/client-error.test.ts
@@ -507,6 +507,27 @@ describe('installChunkErrorHandling', () => {
     expect(global.fetch).toHaveBeenCalledTimes(1);
   });
 
+  // Asserted through the event, not by calling the recovery directly: classification runs first in
+  // this path, and a value whose string conversion throws used to stop the handler there — leaving
+  // the customer on the broken page while a direct unit test of the recovery still passed.
+  it('recovers a chunk failure whose value resists string conversion', () => {
+    install();
+
+    const hostile = {
+      name: 'ChunkLoadError',
+      get message(): string {
+        throw new Error('nope');
+      },
+      toString: () => {
+        throw new Error('nope');
+      },
+    };
+
+    fire('error', { error: hostile });
+
+    expect(reload).toHaveBeenCalledTimes(1);
+  });
+
   // Some engines deliver only the message, with no error object attached.
   it('recovers a chunk failure carried as a bare message', () => {
     install();
--- a/src/__tests__/client-error.test.ts
+++ b/src/__tests__/client-error.test.ts
@@ -2,7 +2,13 @@
 jest.mock('@dfx.swiss/react', () => ({}));
 jest.mock('src/dto/safe.dto', () => ({}));
 
-import { isChunkLoadError, reloadOnceForChunkError, reportClientError, toErrorFacts } from '../util/client-error';
+import {
+  forgetReportedErrors,
+  isChunkLoadError,
+  reloadOnceForChunkError,
+  reportClientError,
+  toErrorFacts,
+} from '../util/client-error';
 
 jest.mock('src/config/api', () => ({ Api: { url: 'https://api.example.com', version: 'v1' } }));
 jest.mock('src/version', () => ({ REACT_APP_BUILD_ID: '42-99' }));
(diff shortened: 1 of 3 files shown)


TaprootFreak:
Four review passes were needed before this came back clean.

Round 1 found four defects, two of them serious. The global error handler called the recovery without the error, so chunk failures outside the router would still not have had their cached answer replaced — half the fix, silently. And the reload waited on the repair with no bound, so a request that never settles would have withheld the recovery indefinitely: worse than before this existed, when the reload was immediate. It is now bounded by a two-second budget as well as by the request settling, and runs exactly once either way.

Round 2 found that looking up the failing address could throw — a throwing getter, or a value that resists string conversion. Because the reload guard is written first, that would have left the customer with neither a repair nor a reload until it expired.

Round 3 found the same weakness one step earlier, in classification itself, which runs before any of this in the global handler. Notably, the existing test passed while that path was broken, because it called the recovery directly instead of going through the event. The new test asserts through the dispatched event and was verified to fail when the fix is removed. That weakness predates this branch; it is fixed here because it sits directly in the path this change exists to make reliable.

Round 4 found one comment that claimed more than the code does — reading a property cannot be assumed free of side effects. Corrected; no behaviour change.

Full suite: 63 suites, 680 tests, green.

PR #1233, opened 2026-08-01
feat(kyc): explain a staff KYC block and offer to start KYC

TaprootFreak:
## Why

Staff endpoints now answer `403 { "code": "STAFF_KYC_REQUIRED" }` when the role is fine but the
account behind it has not completed an identification. Surfacing that as a raw error would leave
staff looking at a bare "Forbidden resource" with no indication that their own KYC is what unblocks
it — the likely reading is "the tool is broken", not "I need to identify myself".

## What

`useGuardedApi` handles the code centrally, right next to the existing `TFA_REQUIRED` redirect. That
hook is where every staff call already goes through, so this covers all staff screens instead of one
dashboard.

It routes to a new screen that

- states that the **role is unchanged** and the identification is what is missing,
- explains that this is the same process customers go through and is done once,
- offers a **Start KYC** button that begins the regular onboarding via `useKycHelper().start()`,
- and a Back button.

Translations added for de, fr and it.

## Depends on

The API side that emits the code is a separate PR in the api repository. Until it is deployed this
screen is simply never reached — the hook only reacts to the new code, so nothing changes for any
existing flow.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,6 +22,7 @@ const SellScreen = lazy(() => import('./screens/sell.screen'));
 const SwapScreen = lazy(() => import('./screens/swap.screen'));
 const AccountScreen = lazy(() => import('./screens/account.screen'));
 const SettingsScreen = lazy(() => import('./screens/settings.screen'));
+const StaffKycRequiredScreen = lazy(() => import('./screens/staff-kyc-required.screen'));
 const BuyFailureScreen = lazy(() => import('./screens/buy-failure.screen'));
 const BuyInfoScreen = lazy(() => import('./screens/buy-info.screen'));
 const BuySuccessScreen = lazy(() => import('./screens/buy-success.screen'));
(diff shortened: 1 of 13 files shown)


TaprootFreak:
Three review passes were needed before both independent reviews came back clean. CI is green on the
current head. What changed along the way:

**Reach.** The first version handled the new code centrally in `useGuardedApi`, and the comment claimed
that covered every staff screen. It did not: `hooks/dashboard.hook.ts` and six staff screens called
`useApi()` directly and would have kept showing a raw 403 — the exact problem this PR exists to fix.
The same gap also meant those screens never got the existing `TFA_REQUIRED` redirect, which that hook
has required all along, so the fix repairs both. All seven now take their `call` from the guarded hook;
each used only `call`, so it is a one-to-one swap.

**Back-bounce.** Navigating here stored the blocked path as the redirect target, so "Back" returned to
a screen that answers 403 again and bounced the user straight back. Unlike the 2FA case, which resolves
in one step, this screen stays blocked until an identification is completed. The redirect is no longer
stored, "Back" goes to the account page, and the stored path is cleared explicitly — dropping
`goBack()` also dropped the only place that cleared it, which would otherwise have misdirected a later
consumer of that single slot.

**Comment accuracy.** The claim that every staff call goes through the hook is now limited to what is
actually guaranteed: staff data hooks and screens do; individual SDK calls such as `useKyc().getFile`
in the compliance screens do not and still surface the error inline.

Note on urgency: the API side is already live, so staff currently see the correct reason as a raw error
with no way to act on it. This PR turns that into an explanation plus a button that starts the process.

PR #1238, opened 2026-08-02
Report which account a client error happened to

### LINKED ISSUE #1237: Client error reports carry no account id, so a reported failure cannot be tied to the customer who hit it
## What happens

A client error report carries `message`, `type`, `stack`, `route` and `version` — but nothing that identifies the account that hit the failure, even when the customer is signed in and the app already knows it.

The resulting log line looks like this:

```
Client error: client="dfx-services" route="/buy" version="…" userAgent="…" type="ChunkLoadError" message="Loading chunk 2688 failed. …"
```

So the record answers *what* broke and *where*, but never *who* it happened to.

## Why it matters

When a customer reports "I can't buy anything, it keeps showing something went wrong", there is no way to look up whether that customer's failure is among the recorded ones — nor how often it hit them, nor whether it stopped.

This is not hypothetical. A support report was checked against the logs today: the failures were in there, but tying them to the reporting customer took a chain of indirect evidence — correlating the time the support entry was updated, resolving client IPs out of trace data, and observing that the customer produced no purchase requests at all during the affected window because the page never loaded far enough. The conclusion stayed circumstantial, and it only worked because the incident happened to be recent and narrow. For anything older, or during a busy period, the same question cannot be answered at all.

An account id in the report turns that reconstruction into a single lookup.

## Proposed change

**1. API — accept and record the field (must come first)**

`CreateClientErrorDto` gains an optional numeric account id, and `ClientErrorService` logs it as its own `key=value` field so it stays queryable:

```
Client error: client="…" account="123456" route="/buy" …
```

This step is a prerequisite, not a nicety: the global validation pipe runs with `whitelist: true`, so an unknown property is stripped silently. Sending the field before the API accepts it would look like it works and record nothing.

**2. Frontend — send it when a session exists**

`reportClientError` adds the identifier the session already exposes (`useUserContext().user?.accountId`), omitting the field entirely when nobody is signed in. Reporting must stay fire-and-forget and must not start depending on user context being loaded — a failure early in startup is exactly the case this reporting exists to catch, and it has to keep working there.

## Constraints worth stating up front

- **The value is client-supplied, so it is a diagnostic hint and nothing more.** The endpoint is deliberately unauthenticated ("the errors worth catching happen before or without a session"), so anyone can post any number. It must never be used for authorization, ownership checks, or anything but correlation — and whoever reads these logs should know that.
- **Optional by definition.** Errors before sign-in are a large part of what this catches; those reports simply carry no account.
- **Keep the existing field limits and sanitizing.** Validate as a number so a free-text value cannot ride in on this field.

## Related

The failure class that prompted this is #1226 (asset misses served as the app shell and cached immutably), which is about the cause. This issue is only about being able to answer who was affected.

TaprootFreak:
Closes #1237 (frontend half).

Depends on DFXswiss/api#4594 — that one must be merged and deployed first. The global validation
pipe strips a property the DTO does not declare, so sending this field before then would look like
it works and record nothing.

## What

A client error report now carries the signed-in account, so a recorded failure can be tied to the
customer who calls support about it. Omitted entirely when nobody is signed in.

## How

`reportClientError` takes the account as an argument and stays a plain function that reads no
React context: it also runs from the window listeners installed at startup, and a failure during
startup — before any session is loaded — is exactly what this reporting exists to catch. Making it
depend on user context being loaded would lose that case.

The error screen passes the account, because that screen belongs to the tree that failed. Keeping
the value in the reporting module instead would have made it shared state: a module is evaluated
once, while `Main.widget.tsx` and `Main.lib.tsx` render one `App` per embedded instance.

The account is part of the deduplication signature for the same reason. The same failure under two
accounts is two customers, and dropping the second would hide exactly the one this reporting exists
to find. That signature is now composed with `JSON.stringify` rather than joined by a separator a
message is free to contain — otherwise two different failures could produce the same signature and
the second would be dropped as a repeat.

The value is left out unless it has the shape of an account id: a safe integer greater than zero,
matching what the endpoint accepts. Losing a report over the field that only helps to find it would
defeat the point of sending it.

## What this value is not

The ingest endpoint is unauthenticated on purpose, so this is a correlation hint that anyone could
post — never an ownership or authorization signal. The API side documents the same.

## What it does not cover

A failure that happens before the session has loaded is reported without an account: the report
goes out as soon as the failure is known, and waiting for a session would risk losing it.

So is a chunk failure that reaches the window listeners rather than the router — including a failed
dynamic import from a click handler, where the customer may well be signed in. Covering that would
mean handing the reporter a session getter from the standalone entry point, evaluated at the time
of the event. It is deliberately left out here: the listeners are installed before any session
exists, and the review that raised it agreed the obvious alternative — keeping the account in the
reporting module — is the design this PR already rejected, because the embedded builds share that
module across instances. Worth doing as a follow-up if the gap matters in practice.

## Tests

- the error screen reports the account of the customer who hit the failure, and reports without one
  when nobody is signed in
- the account is sent, and left out for null, zero, a negative number and one past the safe integer
  range
- the same failure is reported again for a different account, and only once for the same one
- two failures whose values contain the separator stay apart, and an empty error type stays apart
  from a missing one
- a report still goes out when the thrown value carries a name, message or stack that is not text,
  a message that throws when it is read, or a `toString` that throws — the last of these used to
  cost the report entirely, which the old test did not notice because it only asserted that nothing
  escaped
- a value that cannot even be classified still yields its message, so a chunk failure hidden behind
  an unreadable field keeps both its report and the reload that recovers the customer
- `toErrorFacts` converts the fields it can and drops the ones it cannot, and reads a message off a
  thrown object that is not an Error — previously reported as `[object Object]`

`npm run lint`, `npm run test` (691 tests, 63 suites), `npm run build:dev` and `npm run widget:dev`
all pass locally.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,12 @@
-import { DfxContextProvider, PaymentRoutesContextProvider, SupportChatContextProvider } from '@dfx.swiss/react';
+import {
+  DfxContextProvider,
+  PaymentRoutesContextProvider,
+  SupportChatContextProvider,
+  useUserContext,
+} from '@dfx.swiss/react';
 import { SpinnerSize, StyledLoadingSpinner } from '@dfx.swiss/react-components';
 import { Router } from '@remix-run/router';
-import { Suspense, lazy, useRef } from 'react';
+import { Suspense, lazy, useEffect, useRef } from 'react';
 import { LoaderFunctionArgs, Outlet, RouteObject, RouterProvider, redirect } from 'react-router-dom';
 import { LayoutWrapper } from './components/layout-wrapper';
 import { AppHandlingContextProvider, AppParams, CloseMessageData } from './contexts/app-handling.context';
@@ -17,6 +22,7 @@ import ComplianceUserScreen from './screens/compliance-user.screen';
 import ErrorScreen from './screens/error.screen';
 import HomeScreen from './screens/home.screen';
 import { setupLanguages } from './translations';
+import { setReportedAccount } from './util/client-error';
 
 const SellScreen = lazy(() => import('./screens/sell.screen'));
 const SwapScreen = lazy(() => import('./screens/swap.screen'));
@@ -607,6 +613,21 @@ interface AppProps {
   params?: WidgetParams;
 }
 
+// Hands the signed-in account to the error reporter, so a recorded failure can be tied to the
+// customer who calls support about it. A component rather than a call inside reportClientError:
+// reporting also runs from window listeners and from failures during startup, where there is no
+// context to read. Rendered ahead of the router, so the account is in place before any screen can
+// fail.
+function ClientErrorAccount(): null {
+  const { user } = useUserContext();
+
+  useEffect(() => {
+    setReportedAccount(user?.accountId);
+  }, [user?.accountId]);
+
+  return null;
+}
+
 function App({ routerFactory, params }: AppProps) {
   const routerRef = useRef<Router>();
   if (!routerRef.current) {
@@ -624,6 +645,7 @@ function App({ routerFactory, params }: AppProps) {
   return (
     <WindowContextProvider>
       <DfxContextProvider api={{}} data={{}} includePrivateAssets={true}>
+        <ClientErrorAccount />
         <BalanceContextProvider>
           <OrderUIContextProvider>
             <AppHandlingContextProvider
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,7 @@
-import {
-  DfxContextProvider,
-  PaymentRoutesContextProvider,
-  SupportChatContextProvider,
-  useUserContext,
-} from '@dfx.swiss/react';
+import { DfxContextProvider, PaymentRoutesContextProvider, SupportChatContextProvider } from '@dfx.swiss/react';
 import { SpinnerSize, StyledLoadingSpinner } from '@dfx.swiss/react-components';
 import { Router } from '@remix-run/router';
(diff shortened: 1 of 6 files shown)


TaprootFreak:
Fourteen review passes were needed to reach zero findings. Most of them were spent on one theme: reading a value somebody else threw, without letting that reading cost the report.

The substantive findings, in the order they came up:

- The first design kept the account in the reporting module. That module is loaded once per page, while the embedded builds mount one tree per instance — a report could have carried another customer's account. Since a wrong attribution is worse than a missing one on a feature whose only purpose is attribution, the account is now passed in, read by the error screen from the tree that actually failed.
- The deduplication signature ignored the account, so the same failure under two accounts was reported once. It also joined its parts with a separator a message is free to contain, which could merge two different failures into one. It is composed as JSON now, with the account in it.
- Composing that signature could throw on a value with a name that does not serialise — which would have swallowed the whole report. That widened into the real lesson here: classification touches properties before any field is read, so the entire reading is guarded now, and what survives the failure is still used. A chunk failure hidden behind an unreadable field keeps both its report and the reload that recovers the customer.
- `??` where `||` belonged, in three places: an empty message suppressed the fallback that would have named the failure.
- A thrown object carrying a message was reported as `[object Object]`; it now yields the message. That is a plain gain for debugging, unrelated to the account.
- A number of comments asserted more than the code holds — about other repositories, about libraries, about guarantees the endpoint never makes. Shortened rather than reworded again.

Two decisions are deliberate and documented in the description: a failure reported before the session has loaded carries no account, and the window listeners report without one. Both were weighed against the alternative and lost to it.

Issue #1226, opened 2026-07-31
SPA fallback serves the app shell for missing /static/* assets, and it is cached immutably for a year


davidleomay:
## What happens

A request for a `/static/*` asset that does not exist returns the app shell, with a 200 and a one-year immutable cache header:

```
$ curl -sI https://app.dfx.swiss/static/js/main.deadbeef.js
HTTP/2 200
content-type: text/html; charset=utf-8
cache-control: public, max-age=31536000, immutable
x-content-type-options: nosniff
```

Three things compound here:

1. **`200` with `text/html` where JavaScript was expected.** `public/_redirects` has `/*  /index.html  200`, so any asset miss falls through to the shell. With `x-content-type-options: nosniff` the browser is guaranteed to refuse to execute it.
2. **The wrong content inherits the asset cache policy.** The `/static/*` rule in `public/_headers` matches on *path*, not on what is actually served, so the shell comes back as `max-age=31536000, immutable`.
3. **That makes it permanent per client.** Once a browser requests a stale hash even once, that URL is pinned to HTML in its cache for a year. A reload does not clear it; only clearing site data does.

If the file that misses is the main bundle, React never mounts and the page sits on the `Loading ...` placeholder from `public/index.html`. If it is a lazy route chunk, it surfaces as a throw into the router `errorElement`.

## Why it is invisible

The response is a `200`, so it does not appear as an error in Cloudflare traffic analytics, and the dashboard cannot filter on response content-type. There is also no client-side error reporting — `src/screens/error.screen.tsx` never calls `useRouteError()`, so nothing is logged when the boundary fires.

## Proposed fix

Stop the SPA fallback swallowing asset misses, so a missing asset is an honest 404:

```
/static/*   /404.html      404
/*          /index.html    200
```

This also makes `reloadOnceForChunkError()` in `src/index.tsx` usable: a real 404 raises a `ChunkLoadError` that the existing guard can catch and self-heal from, which a 200-with-HTML never can.

Worth considering alongside it:

- Drop `immutable` from `/static/*`, so a bad response cannot pin itself for a year.
- Have `ErrorScreen` call `useRouteError()` and log/report it.
[...]

PR #1244, opened 2026-08-02
Stop a wrong asset response from outliving the deploy that caused it

TaprootFreak:
## What happened

A request for a missing `/static/*` asset was answered by the SPA fallback with the app shell — HTML, status 200. The header rule for `/static/*` matches on the **path**, not on what was actually served, so that HTML inherited `public, max-age=31536000, immutable`. Since `immutable` forbids revalidation, the wrong answer stayed servable for a year in every cache that had fetched it.

Customers landed on the error screen instead of the buy page. One reported case sat there for roughly 40 minutes on a fixed-line connection — not a flaky mobile link, a cached wrong answer. Clearing it required purging the edge by hand.

Measured before that purge: `ChunkLoadError` on `/buy` was the largest single error class in the backend log at 20–42 per hour. Afterwards it fell to 1 per hour, then to 0.

## Why the existing 404 fix was not enough

Missing assets already answer a real 404 (#1230). That stops **new** poisoning — it does not remove a single **existing** one. Two URLs demonstrated the persistence directly: both returned `200 text/html` from cache, 33 and 37 hours old, with a year still to run, while the same URLs with a cache-busting query returned `404`. The application was answering correctly; only the stored copies were wrong.

## What this changes

- **Drop `immutable` from `/static/*`.** The long `max-age` stays — filename hashes change with content, so a conditional request is answered `304` in the normal case. What changes is that a user-triggered reload can revalidate at all, which `immutable` prevents. Shared/edge caches keep the year-long lifetime either way; the comment says so plainly rather than implying this fixes them.
- **Route every other non-fingerprinted path through the content check**, not just the obvious ones: `/favicon.ico`, `/logo.png`, `/robots.txt`, `/manifest.json`, `/asset-manifest.json` and `/version.json`. `no-cache` still *stores* a response — it only forces revalidation before reuse — so a later `304` can confirm a wrongly stored app shell under those URLs. `/` and `/index.html` stay out on purpose: they are the fallback.
- **Fail closed on an absent *or empty* content-type.** An ok response carrying neither is treated as the fallback instead of being passed through, and the media type is compared before its parameters so a type that merely mentions `text/html` in a parameter is not mistaken for one.
- **Verify after every production deploy** that an absent asset really answers a bare 404 with exactly the headers the function emits. A run-specific marker (run id and attempt) is polled first so the check runs against a live deploy. The step states what it can and cannot prove: marker and asset are separate requests and could, during propagation, be served by different edge nodes — it is a smoke test that reliably catches a widespread regression, not a per-node proof.
- **Pin the route list in a test** — exact `include` and `exclude`, so an entry can neither be dropped nor added unnoticed.

## Verification

- `npm run lint` clean; `npm test` green (63 suites, 684 tests)
- Deploy step order checked by hand: the marker is written after every build step and before the upload, so it cannot be overwritten
- Seven review passes against the complete diff. Findings resolved along the way included two that would have made the new gate ineffective — one where it would have re-proved the *previous* deploy, one where a transport failure could have counted as a pass — plus a test that passed while testing something other than its name, and several statements in comments and failure messages that promised more than the code establishes.

Related: #1226 (root cause), #1230 (the 404 answer this builds on).

### DIFF
--- a/.github/workflows/prd.yml
+++ b/.github/workflows/prd.yml
@@ -188,10 +188,11 @@ jobs:
           echo "  url: $url"
           if [ -n "$last_error" ]; then
             echo "  last error: $last_error"
+            echo "This means the verify request itself never got a response from the server (a transport failure, not an answer) -- investigate connectivity to app.dfx.swiss or the runner's network before trusting this deploy."
           else
             echo "  last status: $status"
             echo "  last content-type: $content_type"
             echo "  last cache-control: $cache_control"
+            echo "This means an absent /static/* asset is answered with something other than a plain 404 (e.g. the SPA fallback or other HTML) -- the exact failure mode from the cached-fallback incident. Investigate functions/[[path]].js, public/_routes.json and public/_headers before trusting this deploy."
           fi
-          echo "This means an absent /static/* asset is answered with something other than a plain 404 (e.g. the SPA fallback, HTML, or a transport failure) -- the exact failure mode from the cached-fallback incident. Investigate functions/[[path]].js, public/_routes.json and public/_headers before trusting this deploy."
           exit 1
--- a/.github/workflows/prd.yml
+++ b/.github/workflows/prd.yml
@@ -80,10 +80,11 @@ jobs:
       - name: Verify missing assets return 404 (not the SPA fallback)
         run: |
           # functions/[[path]].js + public/_routes.json turn a missing /static/* asset into a
-          # plain 404 instead of the SPA fallback (index.html, status 200) that used to get
-          # cached under the asset's own URL for a year (see public/_headers). This step is a
-          # smoke test, run after every PRD deploy, for exactly that regression -- see below
-          # for what it actually proves and what it does not.
+          # plain 404 instead of the SPA fallback (index.html, status 200) that used to be
+          # eligible to stay cached under the asset's own URL for up to a year, until evicted or
+          # cleared (see public/_headers). This step is a smoke test, run after every PRD deploy,
+          # for exactly that regression -- see below for what it actually proves and what it does
+          # not.
           #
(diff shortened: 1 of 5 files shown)


TaprootFreak:
Seven review passes against the complete `develop...HEAD` diff, one on conformity and one on logic each round, iterating until neither returned a defect. All checks are green on `ff1a36d`.

The passes were worth it — several findings would have shipped a fix that only looked like one:

- **The deploy gate proved nothing.** The randomised asset URL varied the *path*, not the *version* answering it. Because the previous deploy already answered missing assets correctly, the check went green without ever exercising the build it was meant to guard. A run-specific marker is now polled first.
- **A transport failure could count as a pass.** `|| true` on the probe turned an aborted request into a success once headers had already arrived.
- **The fail-closed claim only held half the time.** An absent content-type was caught, an empty one was passed through as a valid asset.
- **Two more paths carried the same assumption.** `/robots.txt`, then `/version.json` — the latter is written by the build and expected by an e2e test, so a missing file would have been served as the app shell under that URL. The pinned route test would have locked that gap in as correct.
- **A test passed while testing something else.** The empty-content-type case cannot be built through the `Response` constructor: the platform discards `''` and substitutes `text/plain`. The header has to be emptied after construction.

The recurring theme, and the reason the later passes kept finding things, was not broken logic but **statements that promised more than the code establishes** — in comments, in failure messages, once in a test name. Examples corrected here: a cached wrong answer described as removable "only by purging" (eviction and expiry also remove it), a marker loop reporting "deploy did not become active" when all it observed was an unmatched marker, and a comment promising a commit hash where the build writes `unknown` if git metadata is missing.

The last pass returned a single wording defect, fixed in `ff1a36d`; that one-line comment change was not re-reviewed, since the logic dimension had already returned zero findings and everything after pass five was wording without behavioural effect.

PR #1246, opened 2026-08-03
fix(kyc): stop offering KYC start to Compliance accounts on the staff clearance screen

Blume1977:
## What changed

- The `staff-kyc-required` screen no longer offers "Start KYC" to Compliance-role accounts. For them it renders an instruction to contact an administrator instead; the "Back" button stays.
- New translation key in de/fr/it.

## Why

The API rejects self-service KYC for Compliance-role accounts outright (`KYC not allowed for compliance accounts`, api#3577). For such accounts the start button is a guaranteed dead end: `/staff-kyc-required` → `/kyc` → generic red error with the raw API message. The account is instead cleared by an operator through the reviewed backfill path (`docs/staff-kyc-clearance.md` in the api repo, most recently api#4596).

The rule itself stays in the API; the screen only uses the JWT role to render the instruction matching the API's answer, same as the existing role-based rendering in `guard.hook.ts` and `compliance-user.screen.tsx`.

## Checks

- `npm run lint` clean (`--max-warnings 0`)
- `npm run test`: 63 suites / 684 tests pass (includes the translation-completeness suites)
- No API contract change; UI-only

### DIFF
--- a/src/screens/staff-kyc-required.screen.tsx
+++ b/src/screens/staff-kyc-required.screen.tsx
@@ -1,3 +1,4 @@
+import { useAuthContext, UserRole } from '@dfx.swiss/react';
 import {
   StyledButton,
   StyledButtonColor,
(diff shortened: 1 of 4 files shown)


Blume1977:
Closing in favour of DFXswiss/api#4598: that PR removes the dead end at its source (Compliance accounts can run KYC and the owner-initiated mail merge again), so the screen no longer needs to hide the button and send people to an administrator.

Issue #1251, opened 2026-08-03
Limit-request decision form: save re-enables before reload (duplicate side effects), amount semantics not enforced


TaprootFreak:
Follow-up to #1247. Three client-side hardening points from review; the server-side counterparts are tracked in DFXswiss/api#4619.

## 1. Success window: the decision can be submitted twice

`limit-request-decision-form.tsx` (`handleSubmit`): on success the form calls `setIsSaving(false)` and `onDecided()`, but sets `recordedDecision` only in the **failure** branch (`completedSteps.includes('limitRequest')`). `onDecided` is `() => void` and the parent reload is not awaited, so until the reload lands, "Save decision" is enabled again. A second click repeats the deposit-limit PUT and the report POST (a duplicate file record; see the collision/idempotency notes in DFXswiss/api#4619) and then fails on the already-final decision. One-line fix: `setRecordedDecision(decision)` in the success branch as well (or await the reload without re-enabling on its failure). The existing double-submit test covers only the in-flight promise, not this window.

## 2. Amount semantics are not decision-dependent

The amount field is prefilled with the requested limit and validated only as a positive integer. `PartiallyAccepted` can therefore be saved with the full requested amount (prefill untouched) or even more than requested, and `Accepted` with an amount below the request. Suggested rules: `Accepted` ⇒ amount equals the requested limit; `PartiallyAccepted` ⇒ strictly below the requested limit (and above the current one where known); clear or guard the prefill when switching to `PartiallyAccepted`.

## 3. `failedStep` is not shown to the clerk

On failure the form shows the error message and the steps that already landed ("already applied: …"), which satisfies the intent, but the `failedStep` value returned by the hook is discarded. Showing it explicitly ("failed at: report") would spare the clerk deriving it from the sequence order.

Related, documented behavior worth revisiting together with the api-side idempotency work: a retry after a failed decision step re-posts the report, and the earlier file stays in the account's records ("superseded" only semantically, not physically).

TaprootFreak:
Two additions from the follow-up review of #1252 (both pre-existing in the merged #1247 state, out of that PR's scope):

[...]

PR #1252, opened 2026-08-03
34304637 - Harden the limit request decision flow in the compliance tool

TaprootFreak:
Addresses the in-scope items of #1251 (the pre-existing observations recorded there stay open). Follow-up to #1247.

## What changed

- **The decision call now carries the deposit limit.** `decideLimitRequest` sends `grantedDepositLimit` (granting decisions only) with `PUT limitRequest/{id}`; the API raises `userData.depositLimit` itself, behind its own finality check on the request. The separate `PUT userData/{id} { depositLimit }` step is gone. This closes the race where a second, later-clicking clerk could still raise the limit on a request another session had already finalized — the sequence is now report → decision (incl. limit) → note, and the step type narrows to `'report' | 'limitRequest' | 'log'`.
- **The form locks immediately after a successful decision.** `recordedDecision` is now also set in the success branch, so the form drops into note-only mode before the parent reloads — a second click cannot re-fire the decision sequence (and with it a duplicate report file).
- **The failed step is shown.** The error hint now names `failed at: <step>` next to the existing `already applied: …` list.
- **Amounts are validated per decision.** `Accepted` fixes the field to the requested amount (input disabled); `PartiallyAccepted` clears the prefill and requires a whole-number amount strictly below the requested limit and, where known, strictly above the current one, with an inline hint.

## Release order

Requires the api-side counterpart (new optional `grantedDepositLimit` on the limit-request update endpoint) to be **released first** — pair PR: DFXswiss/api#4627. Until then this client would fail loudly on the decision step, leaving the request undecided and retryable (same fail-closed ordering as #1247).

## Checks

- `limit-request-decision.hook.test.ts`: new-sequence call shapes (no `userData` call; `grantedDepositLimit` only on granting decisions), failure positions with correct `failedStep`/`completedSteps`.
- `limit-request-decision.test.tsx`: regression test that the form is in note-only mode right after success (before the parent reload) and a further click does not re-trigger the decision; `failedStep` rendering; per-decision amount validation including the cleared prefill on `PartiallyAccepted`.

### DIFF
--- a/src/__tests__/limit-request-decision.hook.test.ts
+++ b/src/__tests__/limit-request-decision.hook.test.ts
@@ -55,7 +55,7 @@ describe('useCompliance().fileLimitRequestNote', () => {
       clerk: 'JR',
       decision: 'Accepted',
       comment: 'Beleg nachgereicht',
-      attachment: { data: 'data:application/pdf;base64,QQ==', name: 'Kaufvertrag.pdf' },
+      attachment: { data: '[PDF attachment removed — 0 KB]', name: 'Kaufvertrag.pdf' },
     });
 
     expect(outcome).toEqual({ success: true, completedSteps: ['log'] });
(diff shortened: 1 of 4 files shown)


TaprootFreak:
Review complete: 2 full passes (conformity + logic, run in parallel) until zero findings. Pass 1 flagged the attachment fixtures, fixed in b0b0d9a0; pass 2 came back clean on the full diff. Pre-existing observations out of this PR's scope are recorded in #1251 and DFXswiss/api#4619.

PR #1254, opened 2026-08-03
2ac46f80 - Add resume button for stopped transactions in the compliance transactions tab

TaprootFreak:
## What

Adds a **Resume** button next to the existing **Stop** button in the compliance transactions tab. It calls the new backend endpoint `POST /transaction/admin/:id/resume` (DFXswiss/api#4625) and puts a stopped buy-crypto back into automated processing.

## Behaviour

- Shown only for BuyCrypto transactions that are not completed and whose state is `Stopped`
- Confirmation dialog before the call, loading state while running, inline error display on failure
- After a successful resume the transaction detail is reloaded and the surrounding view refreshes

## Changes

- `compliance.hook.ts`: new `resumeTransaction(transactionId)`
- `transactions-tab.tsx`: resume state/handler/button/dialog, mirroring the stop flow
- Prop `onStopped` renamed to `onStatusChanged` (it now fires after stop, resume and chargeback)
- `e2e/compliance.spec.ts`: new visual test covering the resume button and its confirmation dialog, plus the two baselines (the test cancels the dialog, so it never mutates server state). It needs a local buy-crypto with status `Stopped` / `amlCheck` `Pass` — the standard test data has none, see the comment above `TEST_STOPPED_TX_UID`.
- `e2e/test-wallet.ts`: EVM credentials now fetch the sign message from `GET /v1/auth/signMessage` instead of hardcoding it. The API prefixes that message outside production (`[loc]_…`), so the hardcoded text made **every** e2e run fail admin auth with 401 — unrelated to this feature, but it blocks regenerating any baseline.

## Follow-up

Two robustness gaps that this flow inherits from the existing stop flow are tracked in #1256 (not introduced here).

## Depends on

DFXswiss/api#4625 — without the backend endpoint the button returns 404.


TaprootFreak:
Three review passes were needed. What changed along the way:

1. The resume button had no screenshot baseline, which CONTRIBUTING requires for UI changes. Added an e2e test covering the button and its confirmation dialog (it cancels the dialog, so it never mutates server state) plus the two baselines. Producing them surfaced a second problem: `e2e/test-wallet.ts` hardcoded the auth sign message, which the API prefixes outside production — every e2e run failed admin auth with 401. Now fetched from the API, fail-loud.
2. That fix was applied to the EVM path only, while the actively used Bitcoin, Solana and Tron paths kept hardcoding it. All five DFX paths now fetch the message; the two Lightning paths keep their own (different endpoint, no prefix). The then-unused constant was removed. The test's data prerequisite is now documented in full.
3. Final pass: zero blocking findings.

One point both review dimensions raised is tracked in #1256 instead of being fixed here: the action POST and the follow-up detail refresh share a try block, so a failing refresh reports a successful action as "Resume failed". That is unchanged pre-existing behaviour of the stop flow which this change mirrors for consistency, and a repeated resume click is rejected by the backend guard — the impact is a misleading message, not a duplicate action.

Verified locally on every push: `lint`, `tsc -p tsconfig.build.json --noEmit`, the unit suite (725 tests) and the new e2e test against a local stack. All six checks are green and the review bot reported no TypeScript errors.

PR #1255, opened 2026-08-03
b360fbab - Keep the deposit limit consistent with the final limit-request decision

TaprootFreak:
## What changed

Follow-ups to #1247, addressing the review findings on the limit-request decision flow. The invariant this PR enforces client-side: **a finally stored decision must leave a deposit limit that matches it.**

- **Consistency after a partial failure**: when a first attempt already raised the customer's deposit limit (step 1 succeeded) and a later step failed, switching to a non-granting decision no longer leaves the raised limit behind silently. The form passes `revertDepositLimitTo` with the pre-attempt value, step 1 of the retry restores it, and the form announces the restore before saving. Without this, a rejection saved after such a partial failure would become final while the customer kept the raised limit — a state nothing re-checks afterwards.
- **Cumulative step tracking**: the record of an already raised limit survives later failed attempts (a retry that fails before its first step reports `completedSteps: []`, which previously overwrote the knowledge) and early returns such as an unreadable document; it is cleared only on a fully successful save.
- **Fail-closed when the revert target is unknown**: if a revert would be needed but the original limit is not known to the form, the save is refused with an error instead of silently storing a rejection while the raised limit stays.
- **Hook guards its own invariant**: `decideLimitRequest` refuses a granting decision unless `grantedLimit` is a positive integer (`0`, negatives, decimals and `NaN` fail before any API call), instead of relying on the form validation alone.
- **Test hardening**: the double-submit test really submits twice while the first call is pending; the decimal-amount test asserts that no call happens; the revert path, the failed-attempt chain and the unknown-original guard are covered.
- **i18n**: all visible form texts go through `translate('screens/compliance', …)` like the surrounding compliance components. The three decision option labels deliberately stay untranslated: they mirror the API enum values, like the call-queue outcome options.

## Documented limit

The client cannot carry this invariant across a page reload: the transient step state is gone, and re-fetched account data already shows the raised value. The robust fix is server-side (atomic decision endpoint or a consistency check) — tracked in DFXswiss/api#4630. This PR still closes every client-reachable path and is strictly safer than the current develop state, which has all of these gaps open.

## Tests

`limit-request-decision.hook.test.ts` and `limit-request-decision.test.tsx` extended accordingly (44 tests across the two files). Full local gate run green: tests, ESLint (`--max-warnings 0`), Prettier on the touched files, `tsc -p tsconfig.build.json --noEmit`.

### DIFF
--- a/src/__tests__/limit-request-decision.hook.test.ts
+++ b/src/__tests__/limit-request-decision.hook.test.ts
@@ -204,7 +204,7 @@ describe('useCompliance().decideLimitRequest', () => {
     });
 
     expect(outcome).toMatchObject({ success: false, failedStep: 'depositLimit', completedSteps: [] });
-    expect(outcome.message).toContain('needs the limit it grants');
+    expect(outcome.message).toContain('needs a positive integer limit');
     expect(mockCall).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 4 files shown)


TaprootFreak:
Review process: 3 passes (conformity + logic lens each) to zero findings. Pass 1 surfaced 2 conformity and 3 logic findings — the latter all facets of the same root (transient client state cannot fully carry the decision/limit invariant). Fixed in-PR where the client can: cumulative step tracking cleared only on success, fail-closed guard when the revert target is unknown, hardened tests. The remaining reload case is documented in the PR body and tracked server-side in DFXswiss/api#4630. Local gates green on the final head: 44/44 tests, eslint --max-warnings 0, prettier on the touched files, tsc against the build config.

TaprootFreak:
Superseded by #1252, which landed first and removes the root cause this PR was guarding against: the separate `PUT userData` step is gone, the deposit limit now travels inside the decision call and the API applies it atomically behind the finality check — so the partial-failure window (and with it the revert construct, the cumulative step tracking for it, and the unknown-original guard) no longer exists. The remaining delta (the i18n pass over the form) follows in a fresh PR against the current develop.

PR #1257, opened 2026-08-03
b360fbab - Translate the limit-request decision form

TaprootFreak:
## What changed

The remaining delta carried over from #1255 (which #1252 superseded): the limit-request decision form was the only compliance component with fully hardcoded UI texts. All visible texts now go through `translate('screens/compliance', …)` like the surrounding components — headings, field labels, button states, the per-decision amount hints, the failed-step/already-applied error details and both local error fallbacks. Deliberately kept raw: the three decision enum labels (API values, like the call-queue outcome options), clerk names, file names and API-provided error messages.

No logic changes. The form test gains a settings-context stub with real `{{param}}` interpolation so the interpolated assertions stay meaningful.

## Tests

`limit-request-decision.test.tsx` runs unchanged apart from the new stub; full local gates (tests, lint, prettier on the touched files, `tsc -p tsconfig.build.json --noEmit`) before ready.

### DIFF
--- a/src/__tests__/limit-request-decision.test.tsx
+++ b/src/__tests__/limit-request-decision.test.tsx
@@ -24,9 +24,7 @@ jest.mock('src/util/utils', () => ({ toBase64: (file: File) => mockToBase64(file
 jest.mock('src/contexts/settings.context', () => ({
   useSettingsContext: () => ({
     translate: (_ns: string, key: string, params?: Record<string, string | number>) =>
-      params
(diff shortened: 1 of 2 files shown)


TaprootFreak:
Review process: 2 full passes (conformity + logic) plus a final combined confirmation pass to zero findings. Pass 1 surfaced two points — the error-detail parentheses sat outside the translation keys, and the error-detail assertions were substring matches — both fixed (keys carry the full wording, assertions compare byte-exact) and re-verified on the final head. Local gates green: 39/39 tests, eslint --max-warnings 0, prettier on the touched files, tsc against the build config; the render output was reconstructed character-for-character against the previous develop wording.

PR #1268, opened 2026-08-04
fix(handbook): discover documents instead of listing them, and fix slug collisions

TaprootFreak:
Follow-up to #1265, from its closing review. Four findings, each reproduced before being fixed.

## Discovery was not discovery

The README promised the build *finds* artefacts itself — "kein handgepflegtes Mapping". That held for screenshots and nothing else: documents and assets were hardcoded lists.

Adding `docs/whatever.md` would have triggered both workflows through their `docs/**` path filter, rebuilt the handbook, and left the document out. Silently — exactly the failure mode the discovery argument was made against.

It was not theoretical. Switching to a real scan surfaced **two documents that had been missing all along**:

| | before | after |
|---|---|---|
| Documents | 4 (hardcoded) | 6 (scanned) |

The new ones are `infrastructure/Readme.md` and `src/README.md`. The scan excludes `node_modules`, `.git`, `build`, `dist`, `coverage`, `e2e`, any dotted directory, and `docs/handbook` itself.

SVGs under `src/static/assets` are scanned as well. **`public/logo.png` stays curated on purpose** — `public/` holds favicon, manifest, redirects and other files with no business in a handbook, so a scan there would sweep in whatever lands next. The README now states which part is scanned and which is a single curated file, rather than claiming discovery across the board.

## A file that exists was reported missing

Fragment and query were split off the value while it was still HTML-entity-encoded. A numeric character reference such as `&#39;` contains a literal `#`, so the path was cut right there:

```
screenshots/Bank's Guide.png exists; README.md contains
  <img src="screenshots/Bank&#39;s%20Guide.png">
→ handbook integrity check failed: missing screenshots/Bank&
EXIT=1
```

Entities are decoded first now, then the split, then percent-decoding — the order a browser uses. The percent-encoding behaviour that an earlier commit fixed is unaffected.

## Headings could collide

The duplicate counter never registered the suffixed slug it handed out, so a third heading could collide with the first, and the numbering drifted from GitHub's — which matters because the anchors in the sources were authored against GitHub's scheme. Verified against `github-slugger@2.0.0`:

| headings | before | after (= GitHub) |
|---|---|---|
| `Setup-1`, `Setup`, `Setup` | `setup-1`, `setup`, **`setup-1`** ← collision | `setup-1`, `setup`, `setup-2` |
| `A`, `A`, `A-1`, `A-1` | `a`, `a-1`, `a-1`, `a-1-1` | `a`, `a-1`, `a-1-1`, `a-1-2` |

The character class also kept superscripts and fractions, which GitHub drops: `## Fläche in m²` now yields `fläche-in-m`.

## Verification

| Check | Result |
|---|---|
| Build | `178 screenshots, 6 docs, 4 assets` |
| Discovered documents vs. `find` over the repo | identical, 6/6 |
| New `docs/*.md` in a copy | appears automatically; `docs/handbook` stays excluded |
| Entity-encoded reference to an existing file | exit 0 (was exit 1) |
| Reference to a genuinely missing file | still exit 1 |
| Heading sequences | match `github-slugger` exactly |
| Screenshots / `signature-login` / orphan group | 178 / 9 / none |
| Determinism (two builds, `diff -r`) | byte-identical |

No baseline under `e2e/screenshots/` was added, changed or removed; `package.json` and the lockfile are untouched.

### DIFF
--- a/docs/handbook/README.md
+++ b/docs/handbook/README.md
@@ -6,16 +6,18 @@ Docker-Image hinter Basic Auth unter [handbook.app.dfx.swiss](https://handbook.a
 
 ## Wie es funktioniert
 
-Das Assembly-Script `scripts/handbook/build.js` **findet** Artefakte selbst (kein
-handgepflegtes Mapping):
+Das Assembly-Script `scripts/handbook/build.js` **findet** die meisten Artefakte selbst
+(echte Discovery für Screenshots A/B, SVG-Teil von D und Docs E). Die Einzeldatei
+`public/logo.png` ist bewusst kuratiert — `public/` enthält viele unzusammenhängende
+Statik-Dateien, die nicht ins Handbook gehören:
 
 | Quelle | Pfad | Inhalt |
 |--------|------|--------|
 | A | `e2e/screenshots/baseline/*.png` | Playwright Visual-Baselines (flach) |
 | B | `e2e/screenshots/*.png` | Top-Level-Screenshots (nicht rekursiv; `baseline/` und `debug/` ausgenommen) |
 | C | `tailwind.config.js` | Farben und Schriftgrössen |
-| D | `src/static/assets/*`, `public/logo.png` | Logos |
-| E | `README.md`, `CONTRIBUTING.md`, `docs/*.md` | Markdown-Doku (gerendert mit `marked`) |
+| D | `src/static/assets/*.svg` (Verzeichnis-Scan) plus `public/logo.png` (kuratiert, einzelne Datei) | Logos |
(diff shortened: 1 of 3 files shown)


TaprootFreak:
Origin of these changes: the closing review pass over #1265, which ran five review rounds in total. The findings below came out of the fifth and arrived after that PR was already merged, so they land here instead of on the original branch.

The one worth reading twice is the discovery gap: the README promised that the build finds artefacts itself, but documents and assets were hardcoded lists. Switching to a real scan surfaced two documents that had been missing from the handbook all along — `infrastructure/Readme.md` and `src/README.md`. That is the failure mode the discovery argument was made against, present in the feature that argued it.

CI is green on 7 checks, including the handbook image build and container smoke.

PR #1272, opened 2026-08-05
fix(compliance): decide the call-queue AML action from the outcome

Blume1977:
Follow-up to #1225, per review feedback: the AmlCheck selector is offered where there is nothing to decide.

## Problem

#1225 made the AmlCheck action available on every outcome and pre-selected `Reset` on **Completed**. But Completed and Failed already determine what happens to the transaction — the phone-call status written in the same save says it all. Asking the clerk for an AML action there is redundant, invites a wrong pick (`Pass` on a failed call), and is the kind of decision that belongs to the API, not to the tool.

## Change

- **Completed / Failed** → the AmlCheck field is hidden and the action is derived:
  - queue whose AML reason the API keeps out of the recheck cron (`BlockAmlReasons`, i.e. `ManualCheckIpCountryPhone`) → automatic `Reset`,
  - every other queue → nothing is sent; the cron re-evaluates the transaction once the check date is written.
- **Any other outcome** (Unavailable, Suspicious, Repeat) → unchanged: the selector is shown and the chosen action is submitted.
- Switching the outcome clears a previously chosen action, so a stale selection can neither be submitted nor silently reappear.
- The queue predicate lives next to the existing queue mapping in `compliance.hook.ts` as `needsExplicitAmlReset`, so the "which queues does the cron skip" knowledge stays in one place.

`Reset` (never `Pass` or `Fail`) remains the only action the tool sends by itself: it clears `amlCheck` + `amlReason` so the cron re-runs the **full** AML check on the state the call just produced. The decision stays in the API.

Verified against the API for both outcomes: a completed call writes the queue's check date, so the rule that queued the transaction no longer fires and it passes if nothing else is open. A failed call sets `phoneCallStatus = Failed`, and every phone rule in `AmlHelperService` then yields `UserDataFailedCall`, which is mapped to `amlCheck: Fail` / `amlReason: ManualCheckPhoneFailed` — the transaction is failed by the API, exactly as the review asked for.

The queue split is verified the same way: of the call queues, only `ManualCheckIpCountryPhone` carries a reason in `BlockAmlReasons`. `ManualCheckPhone`, `ManualCheckIpPhone` and `ManualCheckExternalAccountPhone` are re-run by the cron on their own, so sending anything for them would only pre-empt the API.

## Fail-closed detail

If the automatic reset is not available for a BuyCrypto on the recheck-blocked queue (KYC not on Check, `buyCryptoResetEligible` false), the save is **disabled** and the form explains the prerequisite. Saving anyway would report success and navigate away while the transaction stays pending forever — in practice this is the common state, not the edge case: the queue members typically sit on `kycStatus = Completed` until the clerk sets Check. On cron-re-evaluated queues an ineligible BuyCrypto does not block the save, since nothing is supposed to be sent there anyway.

## Known behavior on old transactions

For the cron-re-evaluated queues, a transaction whose `amlCheck` has been `Pending` for more than 14 days is failed outright by the API (`aml-helper.service.ts`) before the recheck logic runs. A Completed call on such an old transaction therefore ends in Fail, not in a re-evaluation. That is API behavior this PR does not change, just something to be aware of when working the queue backlog.

## Open question (carried over from #1225)

The recheck path depends on `doAmlCheck`, which is gated by `Process.AUTO_AML_CHECK`. Whether that process is enabled in prod is runtime config — needs a confirmation from @davidleomay/@Yannick1712 before merge, since with this PR there is no manual fallback on Completed/Failed anymore.

## Tests

- Completed and Failed: field hidden, `Reset` on the recheck-blocked queue, nothing on a queue the cron handles.
- Open-ended outcomes: field visible, choice submitted, default no change.
- Outcome switch drops a previous selection.
- User-based queue items never send an action.
- Ineligible BuyCrypto on the recheck-blocked queue: save disabled, reason shown, nothing sent; on a cron-re-evaluated queue the save stays available.
- `needsExplicitAmlReset` pinned against the real `CallQueue` enum: true only for `ManualCheckIpCountryPhone`.
- `npm run test` green, `npm run lint` clean.

### DIFF
--- a/src/components/compliance/call-queue/call-queue-outcome-form.tsx
+++ b/src/components/compliance/call-queue/call-queue-outcome-form.tsx
@@ -11,6 +11,11 @@ import {
   useCompliance,
 } from 'src/hooks/compliance.hook';
 
+// Completed and Failed leave nothing to decide: the phone-call status written in the same save
+// already determines the transaction — the API passes it on the check date a completed call writes,
+// and fails it on a failed one (`UserDataFailedCall`).
+const OutcomesImplyingAmlAction: (CallOutcome | '')[] = [CallOutcome.COMPLETED, CallOutcome.FAILED];
+
 interface Props {
   context: CallOutcomeContext;
   availableOutcomes: CallOutcome[];
@@ -36,11 +41,9 @@ export function CallQueueOutcomeForm({ context, availableOutcomes, clerks, onSav
 
   const hasTx = context.txId != null && context.sourceType != null;
   const buyCryptoResetUnavailable = context.sourceType === 'BuyCrypto' && !context.buyCryptoResetEligible;
-  // Completed and Failed leave nothing to decide: the phone-call status written in the same save
-  // already determines the transaction — the API passes it on the check date a completed call
-  // writes, and fails it on a failed one (`UserDataFailedCall`). So the clerk is not asked for an
-  // AML action there. Every other outcome is open-ended and keeps the selector.
-  const outcomeImpliesAmlAction = outcome === CallOutcome.COMPLETED || outcome === CallOutcome.FAILED;
+  // The clerk is not asked for an AML action on those outcomes; every other one is open-ended and
+  // keeps the selector.
+  const outcomeImpliesAmlAction = OutcomesImplyingAmlAction.includes(outcome);
   const showAmlCheck = hasTx && !outcomeImpliesAmlAction;
   const impliedResetUnavailable =
     hasTx && outcomeImpliesAmlAction && needsExplicitAmlReset(context.queue) && buyCryptoResetUnavailable;
--- a/src/components/compliance/call-queue/call-queue-outcome-form.tsx
+++ b/src/components/compliance/call-queue/call-queue-outcome-form.tsx
@@ -7,6 +7,7 @@ import {
   CallOutcome,
   CallOutcomeContext,
   CallOutcomeResult,
+  needsExplicitAmlReset,
   useCompliance,
 } from 'src/hooks/compliance.hook';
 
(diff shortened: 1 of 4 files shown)


src/components/compliance/call-queue/call-queue-outcome-form.tsx
```diff
@@ -34,17 +35,31 @@ export function CallQueueOutcomeForm({ context, availableOutcomes, clerks, onSav
   }, [clerks]);
 
   const hasTx = context.txId != null && context.sourceType != null;
-  const showAmlCheck = hasTx;
   const buyCryptoResetUnavailable = context.sourceType === 'BuyCrypto' && !context.buyCryptoResetEligible;
+  // Completed and Failed leave nothing to decide: the phone-call status written in the same save
+  // already determines the transaction — the API passes it on the check date a completed call
+  // writes, and fails it on a failed one (`UserDataFailedCall`). So the clerk is not asked for an
+  // AML action there. Every other outcome is open-ended and keeps the selector.
+  const outcomeImpliesAmlAction = outcome === CallOutcome.COMPLETED || outcome === CallOutcome.FAILED;
```
Yannick1712: `[CallOutcome.COMPLETED, CallOutcome.FAILED].includes(outcome)`

Blume1977: Done, moved the two outcomes into a module-level list and switched to `includes`. It needs the `(CallOutcome | '')[]` type because the form starts with no outcome selected.

Yannick1712 (approved):

davidleomay:
Queue split and AML reasoning check out against the API, and the tests are real (8 mutations, all caught). Three things:

**The automatic Reset fires for nothing in the queue today.** Pending `ManualCheckIpCountryPhone` is currently 6 buy_cryptos / 0 buy_fiats, and all six users have `kycStatus = Completed`. The reset needs `Check` (`buy-crypto-reset.util.ts:18`), so `impliedAmlAction()` returns `undefined` for every one of them. Intended — the clerk sets KYC to Check first — but it makes the do-nothing path the normal case rather than the exception.

**And that path reports success.** With `buyCryptoResetEligible: false` the save sends no action, returns `success: true`, and `onSaved()` navigates away (`compliance-call-queue-detail.screen.tsx:192`) — verified by test. The grey hint only appears *before* the save. Disabling Save while `impliedResetUnavailable` would close it.

**The `AUTO_AML_CHECK` question from #1225 is still open**, and now matters more: it gates `doAmlCheck` (`buy-crypto-job.service.ts:27`) and there is no longer a manual fallback on Completed/Failed.

Minor: for the non-blocked queues, dropping the Reset means a Completed call on a >14-day-old tx can be auto-failed (`aml-helper.service.ts:668`) instead of re-evaluated. Probably intended, just not in the description.

Blume1977:
All four points addressed in d278c73:

1. **Do-nothing as the normal case / success on no-op:** the save is now blocked while the implied reset is unavailable (`canSubmit` includes `!impliedResetUnavailable`), with the hint upgraded to an explanation of why saving is disabled and what to do (set KYC to Check, reload, save). The block is scoped to the recheck-excluded queue — on cron-re-evaluated queues an ineligible BuyCrypto does not get in the way, since nothing is supposed to be sent there anyway. Both covered by tests.

2. **14-day auto-fail:** documented in the description under "Known behavior on old transactions" — intended API behavior, this PR leaves it untouched.

3. **`AUTO_AML_CHECK`:** carried into the description as an explicit open question — can you confirm whether the process is enabled in prod? With the manual fallback gone on Completed/Failed, I'd like that settled before this leaves draft.

Blume1977:
@davidleomay PR is ready for review — all four points from your comment are addressed (see d278c73 and the updated description). The only open item is the `AUTO_AML_CHECK` prod-config question in the description; a quick confirmation from you or @Yannick1712 would settle it. I can't request a review on this repo (fork PR), hence the ping.

davidleomay (approved):

PR #1273, opened 2026-08-05
Render a missing KYC file date as a gap, not as an invalid date

TaprootFreak:
## Why

[DFXswiss/api#4695](https://github.com/DFXswiss/api/pull/4695) makes `RealUnitKycFileDto.created` optional. A one-off backfill catalogues legacy KYC documents from the Spider era; for a part of them (among others all name checks) the real document date cannot be established. Instead of sending an invented date — it would be the day of the backfill run and would present a years-old document as brand new — the api omits the field.

## What

- `src/dto/realunit-compliance.dto.ts`: `created` becomes optional, with a note on why a document may carry no date.
- `src/screens/realunit-compliance-user.screen.tsx`: the `Created` column of the KYC file list guards the value and falls back to `-`, the same pattern the neighbouring `Birthday` and `Chargeback` columns already use. Without the guard the missing value reaches `formatDate` and the cell renders the literal `Invalid Date` in the RealUnit compliance dossier's file list.
- `src/__tests__/realunit-compliance-kyc-file-date.test.tsx`: renders the dossier with one dated and one undated file. The test uses the real date formatter and was confirmed to fail (`Invalid Date`) without the guard.

## Scope

`RealUnitKycFileDto` is consumed only by the dossier screen and the RealUnit compliance hook. Nothing sorts, filters or exports on this `created` — the dossier ZIP is assembled api-side — so no other call site had to handle the absent value. The `KycFile` type used by the DFX-internal compliance screens is unrelated and untouched.

## Visual baselines

None affected. Per `CONTRIBUTING.md` the Playwright baselines are regenerated when a change affects the UI. This change alters the rendering only when `created` is absent, and the committed dossier baseline (`realunit-compliance-02-dossier`) is taken from a fixture in which every file has a date, so the screenshot is byte-identical. Nothing was regenerated.

## Merge order

Merge together with or after the api PR. Until that one is live the api always sends the field, so this change is a no-op in production.

### DIFF
--- a/src/__tests__/realunit-compliance-kyc-file-date.test.tsx
+++ b/src/__tests__/realunit-compliance-kyc-file-date.test.tsx
@@ -38,13 +38,17 @@ import { render, screen, waitFor, within } from '@testing-library/react';
 import { RealUnitCustomerDetailDto } from 'src/dto/realunit-compliance.dto';
 import RealunitComplianceUserScreen from 'src/screens/realunit-compliance-user.screen';
 
+// Built from local components so the expectations hold in any timezone the suite runs in: the formatter renders in
+// local time, so a UTC literal would fall back to the previous day on negative offsets.
+const FILE_DATE = new Date(2024, 0, 2);
+
 const DOSSIER: RealUnitCustomerDetailDto = {
   id: 7101,
-  created: '2024-01-01T00:00:00.000Z',
+  created: new Date(2024, 0, 1).toISOString(),
   kycStatus: 'Completed',
   checks: {},
   kycFiles: [
-    { uid: 'file-1', type: 'Identification', name: 'passport.pdf', created: '2024-01-02T00:00:00.000Z' },
+    { uid: 'file-1', type: 'Identification', name: 'passport.pdf', created: FILE_DATE.toISOString() },
     // catalogued legacy document — the api could not establish its date and therefore omits the field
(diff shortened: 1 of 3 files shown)


TaprootFreak:
Review summary — two independent passes, one finding, fixed in the follow-up commit.

The finding was in the new test rather than in the change: the fixture used a UTC instant and asserted the formatted Swiss date, while `formatSwissDate` renders in the local timezone. On a machine with a negative UTC offset the assertion fell to the previous day and the test went red against correct production code, which CI would never have shown (the runner is UTC). The fixture is now built from local components, following the convention and the reasoning already recorded in `src/__tests__/utils.test.ts`. Verified green under America/New_York, Europe/Zurich, UTC and Pacific/Auckland, and still red when the guard is removed.

Also checked: this DTO's `created` has exactly one consumer, the dossier table — no sort, filter or export reads it, so an absent value cannot reorder or break anything else. The placeholder follows the pattern the neighbouring columns in the same screen already use. No committed Playwright baseline is affected: the one dossier baseline comes from a fixture in which every file carries a date, so the rendering there is unchanged. `RealUnitCheckEvidenceDto.date` stays non-optional on the api side and is only ever built from an already dated file, so nothing else needs to follow.

Merge this one first or together with DFXswiss/api#4695. It is a no-op before that PR — the field is always present today — and prevents the file list from formatting a missing date into an invalid one afterwards.

PR #1279, opened 2026-08-07
Write down the pull-request, coverage, handbook and SDK rules

TaprootFreak:
## Problem

`CONTRIBUTING.md` described how to run the two test suites and nothing else. Five rules the team applies in review had no written source, so every one of them had to be restated by hand on each pull request — and a rule that only exists in review comments is a rule new contributors cannot follow before their first round.

## Change

Adds the five missing rules to `CONTRIBUTING.md` and leaves the existing testing guidance untouched:

- **A pull request is self-contained.** Findings on your own PR are fixed in that PR; deferring to a follow-up needs an explicit exception granted in writing by the reviewer.
- **Pre-existing bugs are reported as bugs**, with the same rigour and evidence as newly introduced ones. Not fixing one in the same PR routes through that same written exception.
- **100 % coverage on every touched file** — statements, branches, functions and lines — for the set Jest instruments (`src/**/*.{ts,tsx,js,jsx}` minus `src/**/*.d.ts`). Translation JSON, lock files, assets, type declarations and documentation carry no coverage. The per-file command is documented, and so is the fact that CI has no coverage gate: the numbers go into the PR description and the reviewer checks them.
- **Complete handbook coverage**: every changed screen needs a committed Playwright baseline for each visual variant plus an entry in `scripts/handbook/metadata.json`, including the case where the screen has no spec at all yet.
- **API access goes through the SDK**; a missing parameter or endpoint is added in DFXswiss/packages additively, released, then consumed here.

Two corrections fall out of writing this down:

- `docs/handbook/README.md` called the `metadata.json` entry simply "Optional" and described `handbook-check.yaml` as running on every non-draft pull request. It stays optional for the build, is required by review for a changed screen, and the check is path-filtered — below `src/` it covers only `src/static/assets/**`.
- "Unit tests run in CI on every pull request" ignored both the trigger branches (`develop`, `main`) and the release pull request, where `github.head_ref != 'develop'` skips the build job.

The Playwright section is unchanged. The handbook section states explicitly that creating a missing baseline for a screen you touched does not conflict with the existing "regenerate only the screenshots your change actually affects" rule — the first is about coverage that never existed, the second about not sweeping unrelated baselines into the diff.

## Deliberately not changed

- The coverage rule stays at 100 % of the whole touched file. No diff-coverage variant and no ratchet for long-neglected files; the text says plainly what that costs.
- `handbook-check.yaml` keeps its path filter. Extending it would still not verify per-PR coverage — the build has no diff awareness at all — so the document names the gap instead of implying an automated gate that does not exist.

## Verification

- Every factual claim in the text was checked against the source it names: the two `collectCoverageFrom` globs in `package.json`, the trigger branches and skip condition in `.github/workflows/pr.yml`, the path filter in `.github/workflows/handbook-check.yaml`, and the title fallback and orphan warning in `scripts/handbook/build.js`.
- The documented coverage command was run against both screens of the two open payment pull requests and produces the per-file table it promises.
- Docs-only change: no source file is touched, so there is nothing to cover. Both changed files are inside the handbook's Markdown source set, so the handbook build check runs on this PR once it leaves draft.

## Not verified

Nothing beyond the above — the change adds no code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,13 +15,14 @@ next reviewer then finds the same points again and the work is done twice.
 ### Report every bug you find, including pre-existing ones
 
 A defect in code a pull request touches is reported as a bug — with the same
-severity rating and the same evidence — whether the change introduced it or it
-was already there. Age does not make a defect milder, and "pre-existing" is not
-a category that demotes it to an observation or puts it out of scope.
+rigour and the same evidence — whether the change introduced it or it was
+already there. Age does not make a defect milder, and "pre-existing" is not a
+category that demotes it to an observation or puts it out of scope.
 
 Whether a pre-existing bug is fixed in the same pull request is a separate
-decision and has to be stated explicitly. It is never a reason to leave the bug
-unreported.
+decision, and it follows the same route as any other deferral above: the
+reviewer grants the exception in writing on the pull request. It is never a
+reason to leave the bug unreported.
 
 ## Testing
 
@@ -46,6 +47,12 @@ Partial coverage hides exactly what a code review cannot see either: error paths
 guard clauses and the state combinations a screen only reaches in production. If
 a line genuinely cannot be exercised, delete it rather than excluding it from the
 measurement.
+
+CI runs the suite without a coverage gate, so this is a review gate, not an
+automated one: state the per-file numbers in the pull-request description and let
+the reviewer check them against the diff. Most files in this repository are far
+below the mark today, so touching a long-neglected one means bringing that whole
+file up — plan for it rather than discovering it in review.
 
 ### Visual regression tests (Playwright)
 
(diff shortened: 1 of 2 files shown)


TaprootFreak:
Five review passes were needed to reach zero findings.

Pass 1 (7 findings): the severity wording implied a rating scale this repository does not define; the route for not fixing a pre-existing bug was left vague; the coverage rule did not say where it is enforced; the handbook rule did not cover a screen with no spec and appeared to contradict the build guards in `docs/handbook/README.md`. Two further points — the missing Playwright baseline as a CONTRIBUTING violation, and the empty dependency array freezing the mode — were checked and dismissed with evidence.

Pass 2 (4 findings): the claim about `metadata.json` was inverted — the build accepts a missing entry silently and warns on an orphaned one; "every file" could not hold for translations, lock files or documentation; the path-filter note ignored `src/static/assets/**`; and "unit tests run in CI on every pull request" ignored the release pull request.

Pass 3 (3 findings): `collectCoverageFrom` is two globs, not one; `src/translations/index.ts` is instrumented, so "translations" was too broad an exception; `pr.yml` triggers only against `develop` and `main`.

Pass 4 (1 finding): `docs/handbook/README.md` still described the handbook check as running on every non-draft pull request.

Pass 5: zero findings from both angles.

Every factual claim in the final text was verified against the file it names — `package.json`, `pr.yml`, `handbook-check.yaml` and `scripts/handbook/build.js`.

PR #1282, opened 2026-08-07
Stop resolving internal user IDs to display names on the expenses screen

### LINKED ISSUE #1281: Expenses screen hardcodes internal user IDs mapped to display names in a public repository
`src/screens/dashboard-financial-expenses.screen.tsx:8-11` hardcodes a lookup table that maps internal `userDataId` values to human-readable display names, and renders it at line 156:

```tsx
<td className="py-1.5 px-3">{KNOWN_USERS[r.userDataId] ?? r.userDataId}</td>
```

This repository is public. The table therefore publishes a correlation between our internal account identifiers and the parties behind them — one of the two entries is a personal name. Anyone reading the source learns both which internal IDs exist and who they belong to, without any access to the application.

Note what is *not* the problem: the underlying endpoint `dashboard/financial/ref-recipients` is properly guarded (`AuthGuard() + RoleGuard(ADMIN) + UserActiveGuard()` in the API), and it returns only `userDataId`, never a name. The API is doing the right thing. The disclosure is in the source code, which is the one place that does not sit behind that guard.

The map has been in the file since it was created on 2026-03-19 in #978, so it is not a recent regression.

## Suggested fix

Drop `KNOWN_USERS` and render `r.userDataId` directly. The table already falls back to the raw ID for every recipient not in the map, so the unnamed case is the normal case today and the screen stays usable without the change.

If display names on this screen are genuinely wanted, they belong on the API side of the existing admin guard — resolved in the `ref-recipients` response rather than hardcoded in a public frontend. That is a larger change and a separate decision; removing the map does not depend on it.

## Two things for whoever picks this up

- Deleting the map from `HEAD` does not remove it from the git history, and the repository is public. Whether that residue matters is a call for the maintainers — history rewriting on a public repository is disruptive and should not be done as a side effect of this issue.
- Per `CONTRIBUTING.md`, a pull request touching this file has to bring the whole file to 100 % statement, branch, function and line coverage. The code change is two lines; the coverage work on a long-neglected screen is not. Plan for it rather than discovering it in review.

TaprootFreak:
Closes #1281.

The expenses screen resolved internal `userDataId` values to display names through a hardcoded table and rendered the result in the referral-recipients column. This repository is public, so the table published a correlation between internal account identifiers and the parties behind them — readable from the source alone, without any access to the application.

The map is gone and the cell now renders `r.userDataId`. Every recipient outside the table already rendered as a raw identifier, so this is the path the screen took for all but two rows, and nothing else about the screen changes. The endpoint behind it (`dashboard/financial/ref-recipients`, `AuthGuard + RoleGuard(ADMIN) + UserActiveGuard`) never returned a name; the disclosure existed only in the source.

If display names are wanted on this screen, they belong in that response, behind the guard that already protects it — not hardcoded in a public frontend. That is a separate decision and does not block this change.

## Coverage

The screen had no test at all, so this pull request brings the whole file up, as `CONTRIBUTING.md` requires.

```
npm run test -- --coverage --collectCoverageFrom='src/screens/dashboard-financial-expenses.screen.tsx'
```

| `dashboard-financial-expenses.screen.tsx` | 100 | 100 | 100 | 100 |

`src/__tests__/dashboard-financial-expenses.screen.test.tsx` mocks ApexCharts, the SDK session and the screen's hooks; the chart mock records the props, which is the only way to reach the option formatters, since neither they nor the chart components are exported. Five tests cover the logged-out guard path and the spinner, the transition to logged in, the three chart series builders, the recipient table and both sides of the axis formatter's threshold.

Coverage alone says little about whether a test holds anything in place, so I checked it against mutations of the screen. Each of these fails the suite: dropping `Math.round` from the series builders, switching the CHF locale away from `de-CH`, emptying the effect's dependency array, swapping the first two table columns, swapping two chart headings, changing the chart type, and re-introducing an id-to-name lookup for an identifier the fixture uses.

That last one is worth stating precisely: the test can only catch a re-introduced lookup for identifiers it feeds in. It is not a general guard against the map coming back, and no unit-level fixture can be one without committing the identifiers this pull request removes. If a standing guard is wanted, it belongs in a repository-level check and is a separate decision.

Full suite and lint are green (74 suites, 795 tests), and `build:dev` compiles.

## Handbook coverage — exception requested

`CONTRIBUTING.md` addresses this exact case: *"Every screen or flow a pull request changes has to be represented there"*, and *"If the screen you touched has no baseline yet, create one. That is the case this rule exists for."* This screen has no spec, no baseline and no `metadata.json` entry — nor has any other screen under `/dashboard/**` — and this pull request adds none. So the handbook stays incomplete for this screen, and I am asking for that in writing rather than passing over it.

What the rule would cost here:

- The screen is behind `useAdminGuard`, so a spec needs the API, its database and an admin-seeded account running locally.
- A database created by the documented `npm run setup` has no referral payout rows, so the committed baseline would show an empty table. That is still a valid handbook page for a screen that has none today — but it documents nothing about *this* change, because before and after are identical.
- Making it show the change means seeding payouts for exactly the two identifiers this pull request removes, then committing that rendered payout table as a PNG into a public repository that the handbook republishes. That is the data class the change exists to remove.

So the choice is an empty-state page at the cost of standing up the admin stack, or a content-bearing page at the cost of re-committing what #1281 is about. I have taken neither, and the decision is yours: `handbook-check.yaml` does not run on a pull request that only touches `src/**`, and a `metadata.json` entry without a matching screenshot is an orphan the handbook build warns about, so nothing here is enforced automatically.

If you want the empty-state page, say so and I will add the spec, the baseline and the `metadata.json` entry in this pull request rather than in a follow-up.

## Pre-existing bug in this file (reported, not fixed)

The data load is `Promise.all([...]).then(...).finally(...)` with no `.catch`. If either call rejects, the rejection goes unhandled, the loading flag is cleared anyway, and the admin sees three empty charts and an empty table that look exactly like a successful load with no data. `dashboard-financial-history.screen.tsx` has the same gap; `dashboard-financial-liquidity.screen.tsx` already guards it with `.catch(() => undefined)`.

Filed as #1283 so the report outlives this pull request. Fixing it is a behaviour change unrelated to the disclosure this pull request closes, and doing it properly means deciding what a failed load should look like across all three screens — tell me if you want it here instead and I will add it with its test. The new test deliberately does not pin the current failure behaviour, since asserting it would write today's bug into the suite.

## Residue in the git history

Removing the table from `HEAD` does not remove it from the history of a public repository; it has been there since #978 (2026-03-19). Whether that is worth a history rewrite is a maintainer decision and deliberately not part of this pull request.

### DIFF
--- a/src/__tests__/dashboard-financial-expenses.screen.test.tsx
+++ b/src/__tests__/dashboard-financial-expenses.screen.test.tsx
@@ -13,10 +13,12 @@ jest.mock('@dfx.swiss/react-components', () => ({
   StyledLoadingSpinner: () => <div data-testid="loading-spinner" />,
 }));
 
-const mockChartProps: { options: ApexOptions; series: { name: string; data: number[][] }[] }[] = [];
+type ChartProps = { type: string; height: number; options: ApexOptions; series: { name: string; data: number[][] }[] };
+
+const mockChartProps: ChartProps[] = [];
 jest.mock('react-apexcharts', () => ({
   __esModule: true,
-  default: (props: { options: ApexOptions; series: { name: string; data: number[][] }[] }) => {
+  default: (props: ChartProps) => {
     mockChartProps.push(props);
     return <div data-testid="chart" />;
   },
@@ -50,7 +52,7 @@ type Formatter = (val: number) => string;
 
 // The screen renders CHF amounts through toLocaleString('de-CH'). Building the expectation the same
 // way pins the locale and the digits while staying independent of the group separator the runtime's
-// ICU build emits (U+2019 on current Node, U+0027 on older ones).
+// ICU build emits (U+2019 on ICU 71, U+0027 on ICU 78).
 const chf = (value: number): string => `${value.toLocaleString('de-CH')} CHF`;
 
 // Fractional offsets: the first entry rounds down and the second rounds up, so the Math.round calls
@@ -118,6 +120,14 @@ describe('DashboardFinancialExpensesScreen', () => {
     expect(mockGetFinancialChanges).toHaveBeenCalledWith(undefined, true);
     expect(mockGetRefRecipients).toHaveBeenCalledWith();
 
+    // The headings are what tells the three sections apart once the charts themselves are mocked.
+    expect(screen.getAllByRole('heading', { level: 3 }).map((h) => h.textContent)).toEqual([
+      'Referral Expenses (cumulative)',
+      'Referral Recipients',
+      'Binance Expenses (cumulative)',
+      'Blockchain Expenses (cumulative)',
+    ]);
+
     const [ref, binance, blockchain] = mockChartProps.slice(-3);
     const at = timestamps();
 
@@ -138,7 +148,8 @@ describe('DashboardFinancialExpensesScreen', () => {
     expect(ref.options.colors).toEqual(['#ef4444', '#f97316']);
     expect(binance.options.colors).toEqual(['#f97316', '#8b5cf6']);
     expect(blockchain.options.colors).toEqual(['#3b82f6', '#ef4444', '#64748b']);
-    expect(ref.options.chart?.type).toBe('area');
+    // The type prop is what react-apexcharts honours; options.chart.type would be overwritten by it.
+    expect(mockChartProps.slice(-3).map((chart) => chart.type)).toEqual(['area', 'area', 'area']);
   });
 
   it('lists every referral recipient by its raw identifier, in column order', async () => {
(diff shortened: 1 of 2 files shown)


TaprootFreak:
Three review passes to zero findings.

Pass 1 raised the point that 100 % coverage was not the same as holding the screen in place; the second commit fixed that. Pass 2 raised four points — a chart-type assertion that could not fail, nothing pinning the section headings, an inverted ICU note, and a handbook exception argued from the wrong section of `CONTRIBUTING.md` — all addressed in the third commit and in the description. Pass 3 came back clean on both the conformity and the test-quality lens.

The test was judged against mutations of the screen rather than coverage alone: 38 mutants in the final pass, and every one that changes behaviour fails the suite. The survivors are static ApexCharts display options, moving `setIsLoading(false)` out of `.finally` (the rejection path this pull request deliberately does not pin — see #1283), and an id-to-name lookup for an identifier the fixture does not use, which is the limit stated above.

Two decisions are yours: the handbook exception, and whether #1283 belongs in this pull request rather than on its own.

PR #1285, opened 2026-08-07
a70276af - Add compliance screen for chargebacks awaiting manual approval

### LINKED ISSUE #1278: Compliance screen for chargebacks awaiting manual approval
## Problem

Chargebacks that the automatic release job cannot approve on its own currently wait indefinitely, and nobody sees them. Three cron jobs release refunds automatically (for `BuyCrypto`, `BuyFiat` and `BankTxReturn`), each only when every precondition holds — refund target present, creditor data present, customer not blocked, and for bank refunds the account holder name matching the verified customer name. When one of those fails, the case is skipped silently and stays pending, on every subsequent run.

Until recently an external spreadsheet surfaced those cases for a compliance clerk; that path is gone, so today the backlog is invisible and refunds sit unprocessed for weeks.

The API side is being added separately: a new read-only endpoint `GET /v1/support/pending-chargebacks` (role `COMPLIANCE`) returns every chargeback awaiting manual approval, together with the reason(s) the automatic job cannot release it. This issue covers the frontend that makes that list usable.

Note that approving a single case is **already possible today** via the existing screen at `compliance/bank-tx/:id/return` — what is missing is only the overview that tells a clerk *which* cases need attention.

## Scope

Add a compliance list screen for pending chargebacks, following the existing `compliance-recall-list.screen.tsx` pattern.

- **DTO** — `src/dto/chargeback.dto.ts`, modelled on `src/dto/recall.dto.ts`. Mirrors the API response:

  ```ts
  export enum ChargebackBlockReason {
    NAME_MISMATCH = 'NameMismatch',
    MISSING_CREDITOR_DATA = 'MissingCreditorData',
    MISSING_CHARGEBACK_AMOUNT = 'MissingChargebackAmount',
    MISSING_CHARGEBACK_TARGET = 'MissingChargebackTarget',
    USER_NOT_RELEASED = 'UserNotReleased',
  }

  export interface PendingChargebackEntry {
    txId: number;
    uid: string;
    sourceType: 'BuyCrypto' | 'BuyFiat' | 'BankTxReturn';
    entityId: number;
    userDataId: number;
    userName?: string;
    inputAmount?: number;
    inputAsset?: string;
    chargebackAmount?: number;
    chargebackAsset?: string;
    blockReasons: ChargebackBlockReason[];
    requestedDate: Date;
    date: Date;
    // raw facts for the human decision on NAME_MISMATCH
    verifiedName?: string;
    completeName?: string;
    creditorName?: string;
    chargebackDate?: Date;
  }
  ```

- **Hook** — `getPendingChargebacks()` in `src/hooks/compliance.hook.ts`, calling `support/pending-chargebacks` (same shape as the existing `getRecalls()`).

- **Screen** — `src/screens/compliance-chargeback-list.screen.tsx`, guarded by `useComplianceGuard()`. Table columns: requested date, transaction, customer, input amount/asset, chargeback amount/asset, block reason(s). The API already sorts by `requestedDate` ascending, so the longest-waiting customer is on top — keep that order.

- **Row action** — each row links to the existing approval screen at `compliance/bank-tx/{txId}/return`, so a clerk goes from the list straight into the case. Note the route parameter is the **transaction** id, i.e. `txId` from the DTO, not `entityId`.

- **Route** — register under `compliance/pending-chargebacks` in `src/App.tsx`.

- **Sitemap** — add an entry next to the existing `/compliance/recalls` line in `src/screens/sitemap.screen.tsx`.

## Display detail that matters

`blockReasons` is a list and a case can carry several reasons at once. The two groups are not equally urgent and should be visually distinguishable:

- **`NAME_MISMATCH`** is a genuine human decision — someone must judge whether paying out to a differing account holder is acceptable. For exactly this case the API ships `verifiedName`, `completeName` and `creditorName` alongside; **show all three next to each other**, otherwise the clerk cannot make the call without opening the case.
- **`USER_NOT_RELEASED`** means the customer is blocked or flagged for risk. These must not be waved through casually — at the time of writing, one of the pending cases in production carries a suspicious risk status. Consider marking these rows distinctly.
- The remaining reasons (`MISSING_*`) indicate missing upstream data rather than a decision.

`chargebackDate` is a safety indicator: it must always be empty for rows in this list. If it is ever populated, the case was already refunded and something is wrong — worth surfacing rather than hiding.

## Acceptance criteria

- A compliance user can open the list and see every chargeback waiting for manual approval, oldest request first.
- Each row states why the automatic job cannot release the case.
- For name mismatches, all three names are visible without leaving the list.
- Each row leads to the existing approval screen for that transaction.
- Empty state renders cleanly when nothing is pending.
- The screen is reachable from the sitemap.

## Dependency

Needs the API endpoint to be merged and deployed first. The list will return `404`/`401` until then.

TaprootFreak:
## Why

Chargebacks the automatic release jobs cannot promote wait indefinitely, and since the external tracking sheet was retired nobody sees them — the backlog is invisible and refunds sit unprocessed for weeks. The API side merged as DFXswiss/api#4729: `GET /support/pending-chargebacks` (role `COMPLIANCE`) lists every stuck case with the reason(s) the automatic job cannot release it, oldest customer request first.

## What

A compliance list screen making that backlog actionable, following the `compliance-recall-list` pattern:

- **DTO** `src/dto/chargeback.dto.ts` — mirrors the API response (`ChargebackBlockReason`, `PendingChargebackEntry`).
- **Hook** — `getPendingChargebacks()` in `compliance.hook.ts`, same mechanics as `getRecalls()`.
- **Screen** `compliance-chargeback-list.screen.tsx` — guarded by `useComplianceGuard()`; columns: requested date, transaction, customer, input, chargeback, block reasons, names. The API's oldest-first order is kept, no client-side sort. Loading, error and loaded content render mutually exclusively.
- **Row semantics that matter:**
  - `NameMismatch` rows show `verifiedName` / `completeName` / `creditorName` side by side on one line — that is the data the clerk decides on, visible without leaving the list.
  - `UserNotReleased` rows are visually distinct (yellow row, red badge): blocked or risk-flagged customers must not be waved through casually.
  - `chargebackDate` is a safety sentinel and must always be empty here; if it is ever populated, the row turns red with an explicit error marker instead of hiding the inconsistency.
  - Each row links to the existing approval screen at `compliance/bank-tx/{txId}/return` (transaction id, not entity id).
- **Route** `compliance/pending-chargebacks` + sitemap entry.

## Tests & handbook

- Unit tests bring the screen to 100% statement/branch/function/line coverage and pin the hook contract (`support/pending-chargebacks`, GET). The DTO file measures 0% like every sibling DTO file (repo-wide measurement artifact, e.g. `recall.dto.ts`).
- Playwright spec `e2e/compliance-chargeback-list.spec.ts` with synthetic fixtures (real admin auth, mocked endpoint) plus committed `chromium-darwin` baselines for the loaded list (all row variants incl. the sentinel) and the empty state; handbook metadata entry `compliance-chargeback-list`.
- Warning styles use the repo's real Tailwind palette (`dfxRed`/`dfxYellow`) — the default-palette classes do not exist in this config.

## Notes

- Requires the API endpoint to be deployed; until then the list errors (401/404).
- No new dependencies, no non-null assertions (the widget build is strict about them).

Closes #1278


TaprootFreak:
Three review passes to zero findings.

Fixed along the way: the NAME_MISMATCH name triple no longer wraps onto separate lines; loading, error and loaded content now render mutually exclusively (a failed fetch no longer shows the empty-state text next to the error); the warning styles moved onto the palette this repo actually generates — `bg-red-100`/`bg-yellow-50`/`text-red-800` do not exist in our Tailwind config and rendered no color at all (pre-existing uses of those classes elsewhere carry the same dormant issue and are untouched here); unit tests bring the screen to full coverage; a Playwright spec, committed baselines and a handbook metadata entry cover the new screen.

One conformity note, resolved as out of scope: `getPendingChargebacks()` uses the hook file's local `call<T>()` pattern rather than an SDK method. The endpoint is not in the installed SDK, and all 45 existing functions in `compliance.hook.ts` — including `getRecalls()`, which this feature is modeled on — use the identical local pattern. Moving compliance endpoints onto the SDK is the subject of the dedicated refactor series (#1239–#1243); this PR follows its siblings until that lands.

TaprootFreak:
One note from a later read-through: the unit tests mock the API payload with `Date` instances (`compliance-chargeback-list.screen.test.tsx:45`, `compliance-pending-chargebacks.hook.test.ts:59-60`), but over JSON transport these fields arrive as ISO strings. The rendering path handles both today, yet the tests would stay green through a regression that only breaks on string input — worth switching the mocks to ISO strings to match the real response shape.

PR #1286, opened 2026-08-07
d6222b8a - Preselect the only assignment target on every open

TaprootFreak:
## Summary

The assignment form for an unassigned bank transaction preselected the only available target on the first open, but never again afterwards — leaving customers with an empty dropdown and a permanently disabled submit button.

## Root cause

`assignTransaction` fetched the target list once and ran the single-target preselection **inside the fetch branch**:

- First open: list is fetched, `setValue('target', targets[0])` runs, assignment works.
- `submitAssignment` calls `reset()` after a successful assignment.
- Second open: `transactionTargets` is already populated, so the fetch branch — and with it the preselection — is skipped. The field is empty and `disabled={!isValid}` keeps the submit button greyed out.

## Changes

- The preselection now runs on every open, whether the list was just fetched or came from state.
- A failed target fetch surfaces its error and no longer opens an empty, unusable form.
- Unit coverage for `transaction.screen.tsx` brought up as reported below.
- Handbook coverage for the assignment flow added: `e2e/assign-transaction.spec.ts` plus three baselines and a `metadata.json` entry. The spec mocks every API call including the session and user lookups, so the baselines need no running backend.

## Coverage

Per-file numbers for the files this pull request touches that Jest instruments (test files are excluded from the measurement by the CRA preset):

|---|---|---|---|---|
| `src/screens/transaction.screen.tsx` | 99.62 | 99.71 | 100 | 100 |

Before this pull request the same file stood at 39.25 / 27.12 / 30.23 / 41.76.

## Coverage: the one line the merged report still lists

The merged report lists line 905, `if (!tx.id) return` in the receipt handler. It **is** covered: `transaction-list-txinfo.test.tsx` clicks "Open receipt" on a transaction without an id and asserts the fetch never runs. Measuring that suite on its own shows the line covered; it disappears only when the report merges the runs of several suites that each instrument the module. Reported here rather than papered over, and no exception is being requested for it.

Every line that earlier revisions of this description declared as a deviation turned out not to be one. All are fixed rather than excused:

- `if (!transaction) return` in `handleTransactionNavigation` — the value was never used again. Deleted.
- `if (!editTransaction) return` in `submitAssignment` and `if (!transaction?.id) return` in the refund submit — both ids now arrive as parameters. The first also rejected the id `0`, which a test covers.
- `if (txRefs.current)` — `useRef<Record<…>>({})` makes it permanently truthy. Deleted.
- `onSubmit` on both the assignment and the refund `<Form>` — the shipped `@dfx.swiss/react-components` never invokes it: `Form` renders the form element without binding it and forwards the prop only to children with a `name`, and none of `StyledDropdown`, `StyledInput` or `StyledSearchDropdown` reads it. Submitting runs through the button handlers, which are tested. Deleted in both places; the same pattern elsewhere in the repository is a separate question.
- `items={transactionTargets ?? []}` — expressed as a render condition instead.
- The chargeback address description guarded a blockchain the render condition already required, and a test reached that branch with an object whose `toString()` returns `undefined`. Both are gone; a proper test now covers a refund whose transaction has no blockchain.

One regression was caught in review and fixed here: narrowing the chargeback address dropdown on `inputBlockchain` skipped the field entirely for a sell refund without one. `Form` only hands validation rules to controls it renders, so the required address stopped being registered and the form could pass validation without one.

Two earlier commit messages on this branch quote test counts of 978 and 980. Those were inflated by a stray copy of a test file on the build host that counted as an extra suite; the real figure is 936. The messages are left as they are rather than rewriting pushed history.

## Incidental finding

`.env.sample` sets `REACT_APP_API_URL=http://localhost:3000          # Local API`. The trailing comment ends up inside the value, so the app throws `Failed to parse URL` on every request and the documented baseline workflow fails until the line is trimmed. Pre-existing and untouched here.

## Verification

- `lint`, `build:dev` and `widget:dev` pass; 936 tests pass in 80 suites.
- `npx playwright test e2e/assign-transaction.spec.ts --project=chromium` passes against the committed baselines, both freshly generated and on a second run.
- Mutation test: moving the preselection back into the fetch branch turns two of the three new wiring tests red, and reverting restores them.


TaprootFreak:
Seven review passes across two dimensions (conformance and logic); one pass was lost to a tool timeout and repeated, so it does not count.

What the reviews changed, beyond the original fix:

- Six lines this description once declared as unreachable coverage were not deviations at all. Two dead guards were deleted, two ids now arrive as parameters (one of them had silently rejected the id `0`), a `?? []` fallback became a render condition, and the inert `onSubmit` prop came off both forms. The declaration is gone; the rule is met.
- One regression introduced mid-review was caught and fixed: narrowing the chargeback address dropdown on `inputBlockchain` dropped the field entirely for a sell refund without one, which unregistered the required-address rule.
- Several tests were passing for the wrong reason — a button mock that swallowed `disabled`, a search mock that invoked the very callbacks under test, an assertion inside an `if` that could never fail, and a fixture that faked an SDK type to reach a branch. All now assert real behaviour.
- The e2e spec only exercised the first open, which worked before the fix too. It now submits and reopens, which is where the bug actually was, and pins the timezone so the baselines are not tied to one machine.

Remaining minor point filed as #1303.

PR #1288, opened 2026-08-08
Add a full-stack E2E harness and a functional suite covering every screen

TaprootFreak:
## What this adds

A full-stack end-to-end harness under `e2e-stack/`, and a functional test suite that runs on it.

One command brings up the real frontend, the real API and a real Postgres, runs the suite against
them, and tears everything down again:

```
npm run e2e:stack
```

The point is to check a pull request against the actual interplay of the three parts before it
merges, rather than against unit-test mocks on either side. A screen that renders, an endpoint that
answers, and a row that lands in the database are three different claims; until now only the first
two were tested here.

## What is real and what is mocked

Real: Postgres (fresh and ephemeral), the API, the frontend, the browser.

Mocked: every external provider — banks, KYC/AML, exchanges, blockchain nodes, pricing, storage.

That happens on two levels, and the second one is the load-bearing one:

1. The API runs with `ENVIRONMENT=loc` and mocks its own outbound calls.
2. The API sits on a Docker network with `internal: true` and has no route to the internet at all.

The second level matters because the first only covers calls that go through the API's central HTTP
wrapper. Around twenty integrations use vendor SDKs, `graphql-request`, ethers providers or bare
`fetch` and bypass it entirely — two of them fire during boot. Without the network isolation a test
run would reach real third-party systems. The isolation is a security property of this harness, not
an optimisation, and the comments in `compose.yml` say so.

## Coverage

Every route in `src/App.tsx` is covered — 100 of them, across eleven suites — plus cross-cutting
behaviour that does not belong to a single screen: language switching, deep-link parameters and
whether they are stripped from the address bar afterwards, API error handling, session expiry, small
viewports, and the widget entry point (a second entry point that had no end-to-end coverage at all
until now).

The last run: **220 tests green in about seven minutes**.

Role checks assert both halves: that the frontend routes a denied role away, and that the API itself
answers 403 for the same account. Either one alone would pass while the other quietly regressed.

Coverage is enforced rather than tracked by hand. `route-coverage.spec.ts` reads the route
definitions out of `src/App.tsx`, resolves nested paths, and fails when a route is claimed by no
suite or by more than one. Anything the parser cannot resolve — a computed path, a spread, a
shorthand property — is a hard failure rather than a silent omission, because a route it cannot read
is a route that would never need a claim.

Where a screen writes, the test drives the write through the interface and then asserts the row in
Postgres. Signature login and email login both run end to end — the one-time code is read from the
database, because outbound mail is off in this environment.

## No test passes because the defect is there

An earlier revision of this suite contained a pattern worth naming: tests that checked for a known
defect at runtime and skipped themselves when they found it. That is worse than no test — the suite
reports success exactly while the defect is present, and would never go red if someone reintroduced
it later. None of those are left.

What replaced them:

- A known defect a test can still exercise is marked as an expected failure. The test runs, its
  failure is visible in the report, and the moment the defect is fixed the run turns red until the
  marker is removed. That has already happened three times during this work: tests marked as
  expected failures started passing and the markers came off.
- Everything else was a test that did not check what its name claimed — an assertion satisfied by
  the page body, a wait already satisfied on the page it started from, a precondition skipped when a
  seed happened to be empty — and now does.

## Running it

- `npm run e2e:stack` — up, test, tear down
- `npm run e2e:stack:up` / `npm run e2e:stack:down` — leave the stack up to poke at it

Prerequisites: Docker with Compose v2, Node 20, and either the API repository checked out as a
sibling directory or `E2E_API_IMAGE` pointing at a prebuilt image. Details, including the
troubleshooting notes, are in `e2e-stack/README.md`.

## Relation to `e2e/`

The existing suite under `e2e/` is visual-regression testing and deliberately does not run in CI,
because its baselines are platform- and font-dependent. This harness checks function rather than
appearance, so it does run in CI, on every pull request. Both stay.

## Known limits

A handful of paths are held open with a stated reason rather than asserted loosely so they pass:
behaviour that genuinely cannot be reached without a real third party (card payments, on-chain
signing, live Lightning), and the known product defects listed below. Each one names which of the
two it is.

## Findings

Writing the suite turned up issues in the application itself. Two are fixed in the API pull request
that accompanies this one, because without them the API cannot start in an isolated network at all.
The rest are reported separately, and the tests that would catch them are in place as expected
failures — including a `/sell` deep link that makes the screen create bank accounts in a loop and
never price anything (#1295), which blocked the entire sell flow until it was tracked down.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,9 +1,12 @@
+# Applies to every image built from this repository, not just the e2e-stack ones. Anything listed
+# here disappears from the build context of Dockerfile.handbook too — which is how an earlier
+# version of this file, excluding e2e/screenshots, broke the handbook build: that image copies the
+# screenshot baselines in. Check the other Dockerfiles before adding an entry.
 node_modules
 e2e-stack/node_modules
 .git
 build
 dist
-e2e/screenshots
 test-results
 playwright-report
 *.log
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+node_modules
+e2e-stack/node_modules
+.git
+build
+dist
+e2e/screenshots
+test-results
+playwright-report
+*.log
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,3 +10,7 @@ dist
 test-results
 playwright-report
 *.log
+# CI checks the API repository out into the workspace as api-repo/ before building the frontend
+# image from this directory. Without this line the whole of it travels to the Docker daemon as
+# build context on every image build.
+api-repo
--- a/.github/workflows/e2e-stack.yml
+++ b/.github/workflows/e2e-stack.yml
@@ -75,9 +75,12 @@ jobs:
       - name: Run Playwright tests
         # No continue-on-error: a failed test must fail the job. Later steps use
         # `if: always()` so artifact collection and teardown still run.
+        # --env-file is not optional here: up.sh writes the values it resolved into that file, and
+        # a compose run that resolves them differently recreates the API container mid-run.
         run: |
           docker compose \
             -p dfx-e2e-stack \
+            --env-file e2e-stack/.env \
             -f e2e-stack/compose.yml \
             -f e2e-stack/compose.tests.yml \
             run --name dfx-e2e-stack-tests tests
--- a/.github/workflows/e2e-stack.yml
+++ b/.github/workflows/e2e-stack.yml
@@ -80,7 +80,7 @@ jobs:
         run: |
           docker compose \
             -p dfx-e2e-stack \
-            --env-file e2e-stack/.env \
+            --env-file e2e-stack/.env.generated \
             -f e2e-stack/compose.yml \
             -f e2e-stack/compose.tests.yml \
             run --name dfx-e2e-stack-tests tests
--- a/.github/workflows/e2e-stack.yml
+++ b/.github/workflows/e2e-stack.yml
@@ -56,6 +56,15 @@ jobs:
           node-version: '20'
           cache: 'npm'
 
+      # The suite sits outside the repo's own tsconfig and ESLint globs (both scoped to src/),
+      # so nothing else in CI type-checks it. Doing it here costs under a minute and fails on a
+      # type error before the run spends eight on Docker.
+      - name: Type-check the harness
+        working-directory: e2e-stack
+        run: |
+          npm ci
+          npx tsc --noEmit
+
       # run.sh is not used here: its EXIT trap tears the stack down with
       # `docker compose down -v`, which destroys the named volumes that hold
       # Playwright traces, screenshots, videos, and the HTML report before they
--- a/.github/workflows/e2e-stack.yml
+++ b/.github/workflows/e2e-stack.yml
@@ -4,6 +4,13 @@
 # A skipped GitHub Actions check counts as passing, which would silently defeat the
 # merge gate. If you ever consider adding a skip anywhere in this file, justify it in
 # a comment right there.
+#
+# Merge order with DFXswiss/api: the corresponding API-repo workflow checks out
+# DFXswiss/services@develop by default and expects e2e-stack/ to exist there. That is only
+# true once *this* pull request (which introduces e2e-stack/) has been merged to develop.
+# Merge this services PR first, then the API-repo PR. Until then, the API-repo workflow can
+# only be exercised via workflow_dispatch with a services_ref pointing at a branch/commit
+# that already has e2e-stack/.
 
 name: Full-stack E2E
 
--- a/.github/workflows/e2e-stack.yml
+++ b/.github/workflows/e2e-stack.yml
@@ -1,0 +1,95 @@
+# Full-stack E2E: real frontend + API + Postgres against mocked external providers.
+#
+# Do not add a job-level `if:` skip condition (for draft PRs, path filters, or similar).
+# A skipped GitHub Actions check counts as passing, which would silently defeat the
+# merge gate. If you ever consider adding a skip anywhere in this file, justify it in
+# a comment right there.
+
+name: Full-stack E2E
+
+on:
+  pull_request:
+    branches:
+      - main
+      - develop
+  workflow_dispatch:
+    inputs:
+      api_ref:
+        description: 'Git ref of DFXswiss/api to check out and build against'
+        required: false
+        default: 'develop'
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: e2e-stack-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  e2e:
+    name: Full-stack E2E
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - name: Checkout services
+        uses: actions/checkout@v4
+
+      - name: Checkout API
+        uses: actions/checkout@v4
+        with:
+          repository: DFXswiss/api
+          ref: ${{ inputs.api_ref || 'develop' }}
+          path: api-repo
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      # run.sh is not used here: its EXIT trap tears the stack down with
+      # `docker compose down -v`, which destroys the named volumes that hold
+      # Playwright traces, screenshots, videos, and the HTML report before they
+      # can be copied out. Instead we bring the stack up, run tests, copy
+      # artifacts from the still-present named volumes / tests container, and
+      # only then tear down in a final always() step.
+      - name: Bring stack up
+        run: E2E_API_REPO=$GITHUB_WORKSPACE/api-repo bash e2e-stack/scripts/up.sh
+
+      - name: Run Playwright tests
+        # No continue-on-error: a failed test must fail the job. Later steps use
+        # `if: always()` so artifact collection and teardown still run.
+        run: |
+          docker compose \
+            -p dfx-e2e-stack \
+            -f e2e-stack/compose.yml \
+            -f e2e-stack/compose.tests.yml \
+            run --name dfx-e2e-stack-tests tests
+
+      - name: Collect test artifacts
+        if: always()
+        run: |
+          mkdir -p e2e-stack-artifacts
+          # Named volumes back these paths inside the tests container. Bind mounts
+          # are not used. Copy out before down.sh removes the volumes (-v).
+          docker cp dfx-e2e-stack-tests:/work/test-results e2e-stack-artifacts/test-results \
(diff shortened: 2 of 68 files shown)


TaprootFreak:
Four review passes were needed before both reviewers reported no findings.

What they turned up, in order:

1. Infrastructure and shared fixtures — the cleanup order deleted `user_data` before `user`, against the foreign key, so it failed on every run; the wallet-index derivation could restart low and silently reuse a previous run's account; several documented commands and paths did not match what the code actually did.
2. Full diff — role denial was only ever checked as a frontend redirect, never against the API's own answer, even though the fixture option for it existed unused; cleanup failures were collected and then discarded, leaving no trace of a polluted database for whichever spec ran next.
3. The fix commits themselves — a comment asserted a lint rule that does not exist anywhere in this repo.
4. Confirmation — no findings.

Alongside those, the formatting of 23 of 24 spec files was brought in line with the repo's Prettier configuration, in a commit kept separate from the content changes so the diff stays readable.

The last CI run: 203 tests green, 16 held open with a stated reason, in 7.5 minutes.

One note for whoever merges: this has to land before the counterpart in the API repository, which checks this repository out to find the harness.

TaprootFreak:
Review continued after this was first marked ready, and it was worth it — the later passes found
things the earlier ones had not, so the suite has changed substantially since. The final pass
reported no findings.

What came out of it, in order of how much it mattered:

1. **Tests that passed because the defect was present.** Several tests checked for a known defect at
   runtime and skipped themselves when they found it. The suite therefore reported success exactly
   while the defect was there, and would never have gone red if someone reintroduced it. None are
   left: a known defect that a test can still exercise is now marked as an expected failure, so the
   test runs and the run turns red the moment the defect is fixed and the marker outlives it. That
   already happened three times during this work.

2. **A blocked flow that turned out to be a product defect.** The sell tests navigated with a
   `bank-account` deep link parameter. The screen answers that parameter by creating bank accounts
   in a loop — more than thirty times in twenty seconds — and never requests payment information, so
   nothing downstream could work. Without the parameter the whole flow completes, which is why the
   payment panel tests are now hard assertions rather than expected failures. The defect is reported
   as #1295 and has a test of its own.

3. **A global option that hid crashes.** The stack ran the API with unhandled rejections downgraded
   to warnings, which covers every unhandled rejection rather than the one known case, so a genuine
   crash could have passed for a healthy run. The stack now probes the image for the handler that
   makes the downgrade unnecessary and only sets it when that is absent — it disappears by itself.

4. **Three ways past the coverage gate.** A route whose path is a variable, a spread, a shorthand
   property or a reference simply vanished from the set of routes needing a claim; two claims for
   the same route inside one registry file slipped through; and the spec filename a claim names was
   never checked to exist. All three fail loudly now.

5. **Fixtures that deleted rows they had not created**, a cleanup order that removed a parent before
   its child, a role change that handed back a token minted before it, a screen still showing its
   loading spinner counting as rendered, and a session hop satisfied by whatever token happened to
   be in storage rather than the one being handed over.

Not everything suggested was taken. Waiting for every loading indicator to disappear rather than the
first was implemented, measured and reverted: some screens keep a status region for good, so that
version sat out its full timeout on healthy screens, more than doubled the suite's runtime and
failed two tests that had nothing wrong with them. The reasoning is written down where the decision
is.

Current state: **220 tests green in about seven minutes** against a stack built from scratch, and
all checks green here. The handbook check is skipped by its own path filter, as designed.

One note for whoever merges: either repository can go first. The counterpart bootstraps this
harness from this pull request until it reaches the default branch.

TaprootFreak:
@mara-steiner please check

TaprootFreak:
@mara-steiner please check

TaprootFreak:
Thank you — running the stack yourself and probing the isolation from inside the container is more
than I could reasonably have hoped for, and every one of the three main items was right.

## The three change requests

**1. A run with zero tests exited 0.** Fixed. `run.sh` no longer has an optional branch: a missing
`tests` service is a hard failure, and `compose config` keeps its stderr so a broken compose file is
readable rather than silently treated as "no tests here".

**2. The gate checked file existence, not coverage.** Fixed, and by binding claims to what actually
ran rather than by renaming the gate. The browser fixture records every navigation it makes, and the
gate compares that recording against the route list: a claim pointing at a file that never opens the
route no longer satisfies it. Measured before committing to it — of the 100 routes, 99 were already
being navigated to, and the hundredth only because an earlier failure had cut its file short, so the
check costs nothing in coverage that was already there.

The recording is only enforced on a full run (`E2E_FULL_RUN=1`, set by `run.sh` and by both CI
workflows). A filtered run — one spec file while working on it — legitimately visits a fraction of
the routes, and failing there would teach people to ignore this gate; it says so instead of quietly
checking nothing. `CONTRIBUTING.md` now describes exactly this, so the guide and the tooling agree.

**3. `--unhandled-rejections=warn` applied process-wide.** Gone, with nothing in its place. You were
right that the self-disabling probe was the good part and the fallback was broader than the problem.
The probe now refuses an API image whose process error handling can throw while logging, and the
stack fails to start rather than run with every unhandled rejection silenced. Until the guard is on
the API repository's default branch, this workflow builds the API image from the pull request that
carries it — the mirror image of what that repository's workflow does with this one.

## The smaller items

- **`buy.spec.ts`** — the outcome set is gone. Measured against the stack, the API answers
  `No personal IBAN provider available for this currency`, so the test asserts that one named
  outcome: a different error, or a silent success, now fails it. A comment says what to seed to turn
  it into an assertion on a generated IBAN.
- **`factories.ts` success-path fallthrough** — fixed. A 2xx that yields no id now throws instead of
  falling into the SQL fallback and fabricating what the API failed to produce.
- **`retries`** — 0. Your reasoning about a shared database is exactly right.
- **`cross-cutting.spec.ts:729`** — the assertion is unconditional.
- **duplicate `track('deposit')`** — removed; `insertReturningId` already registers it.

## From the second pass

- `CONTRIBUTING.md` and the gate now say the same thing (above).
- The README's port-override instruction is gone rather than wired through: the bundle's API URL is
  what the browser *inside* the tests container uses to reach the loopback forwarder, so deriving it
  from a host debug port would change what the suite runs against. The README now states what
  actually holds.
- The "one deliberate exception" sentence is corrected — 13 raw `fetch` call sites across 8 files.
  The reasoning covers all of them; only the count was wrong.
- `getUserDependentTables().catch(() => [])` is gone: schema discovery failing now fails the cleanup
  rather than silently deleting nothing.
- The four unused page fixtures are removed, `normPath` and the route parser live in one shared file
  (`specs/fixtures/routes.ts`) instead of 13 copies and a lenient private variant, the ~24 non-null
  assertions became named checks, the cache key points at the harness lockfile, and the blank line
  at EOF is gone.
- `E2E_LOOPBACK_PORT` was read by the entrypoint but never passed through — it is now.

## Recorded rather than fixed

Teardown still cannot delete every row: the application writes rows of its own during a test
(`kyc_file`, `buy_crypto`, and the `user`/`user_data` pair), those are not registered, and the
deletes fail on foreign keys. It warns rather than fails, which is your point about the ignored
`{ errors }`. Closing it means registering application-written rows first and only then making the
cleanup fail loudly — doing it the other way round turns a green suite red for reasons unrelated to
the code under test. Filed as #1299 with the observed rows.

The two defects you singled out are tracked outside this file: the `/sell` bank-account deep link as
#1295, and the `/file/:id` access scope with the team through the private channel. Your point that
only the recorded defect is pinned on those routes is fair and worth keeping in mind — an unrelated
regression on the same route would not stand out.

Full run against a stack built from scratch after all of the above, and all checks green here.

TaprootFreak:
@mara-steiner please check

TaprootFreak:
**EN:** The review findings are filed as #1311 and #1312 rather than fixed here, so the harness can merge and the gate-integrity work lands separately.
**DE:** Die Review-Funde sind als #1311 und #1312 erfasst statt hier behoben, damit der Harness mergen kann und die Gate-Integritätsarbeit separat landet.

<details>
<summary>Details</summary>

**#1311 — gate integrity (the three items marked as merge blockers in the review):**
- `forbidOnly` is missing from `playwright.config.ts`, so a committed `test.only` would skip the bulk of the suite and the `coverage-gate` project while CI still reports green.
- The root `.dockerignore` does not exclude `.env`, `**/.env`, `**/.env.*` or `e2e-stack/.env.generated`, and the frontend build context is the repo root — so those files travel to the Docker daemon and into builder layers.
- `cleanupCreatedData` collects delete failures and only warns, while clearing the registry before the deletes; leftover rows in the shared database can make later specs false-green or false-red without failing the file that caused it.

**#1312 — medium items:** route coverage is a global path set rather than bound to the claiming spec file, and the 2xx-without-id hardening exists only on `createLimitRequest`.

What this PR does contain, verified by a full local stack run before pushing (222 passed, 0 failed): the compliance route inventory is complete at 25 paths, `/compliance/pending-chargebacks` and the catch-all `/compliance/*` are claimed and asserted, and `routeMatches` now expands a splat to the rest of the path — previously it escaped `*` literally, which made a catch-all route impossible to satisfy no matter which URL was visited.

</details>

TaprootFreak:
@mara-steiner please re-check — your three gate-integrity findings are accepted as correct, and they are filed rather than fixed here.

**EN:** All three of your blockers are tracked in #1311 and the two medium items in #1312, so this PR is meant to merge with them open — please re-check on that basis and update your review if you agree.
**DE:** Alle drei Blocker sind in #1311 erfasst und die zwei mittleren Punkte in #1312, dieser PR soll also mit diesen offenen Punkten mergen — bitte auf dieser Grundlage erneut prüfen und dein Review entsprechend aktualisieren, falls du einverstanden bist.

<details>
<summary>Details</summary>

Nothing in the code changed since your review of `cc7be2ec` — this is a request to re-check the disposition, not the diff.

**Why they are filed and not fixed:** the decision was to land the harness now and treat gate-integrity hardening as follow-up work, so the E2E stack stops being a moving target for everything else that depends on it. Your three findings are not disputed:

- `forbidOnly` missing — a committed `test.only` would green CI while skipping the suite and the `coverage-gate` project.
- Root `.dockerignore` not excluding `.env` / `**/.env.*` / `e2e-stack/.env.generated`, with the repo root as the frontend build context.
- `cleanupCreatedData` collecting delete failures behind a `console.warn` while clearing the registry before the deletes.

All three are written up in #1311 with your reasoning and the suggested fixes, including the point that successful cleanup must stay silent so the change does not turn the suite red.

**What this PR does contain,** verified by a full local stack run before pushing (222 passed, 0 failed): the compliance route inventory is complete at 25 paths, `/compliance/pending-chargebacks` and the catch-all `/compliance/*` are claimed and asserted, and `routeMatches` expands a splat to the rest of the path — it previously escaped `*` literally, so a catch-all route could never satisfy the gate no matter which URL was visited.

CI on this head is green across all 8 checks.

</details>

PR #1296, opened 2026-08-08
docs: set clipboard-write permission on the Iframe example

Danswar:
## Summary

`clipboard-write` is a Permissions-Policy-gated feature whose default allowlist is `self`. In a cross-origin Iframe — which is what the documented Iframe integration of app.dfx.swiss is — `navigator.clipboard.writeText` therefore rejects unless the embedding page grants the permission via `allow="clipboard-write"` on the Iframe element.

With #1287 the app copies via `navigator.clipboard.writeText` and only falls back to the legacy DOM-based mechanism when the Clipboard API is unavailable or the write rejects. Without this attribute, an Iframe integrator's users take the fallback on every copy in Chromium; with it, they get the native path. The attribute is an upgrade, not a prerequisite — copy buttons keep working without it (the fallback is not policy-gated), which is why the docs sentence recommends rather than requires it, and why it is phrased around browser behavior so it is accurate both before and after #1287 lands. Enforcement of `clipboard-write` delegation is Chromium-specific; Firefox and Safari gate the Clipboard API on user activation instead, so the attribute is a no-op there.

This PR adds the attribute to the Iframe code example and states the recommendation in the Iframe section. It also closes the `<iframe>` tag in the example — `iframe` is not a void element, so the previous self-closing form was invalid HTML that would swallow any markup an integrator pasted after it.

The Web Component and React Component integrations load into the host page's own document and need no attribute.

## Test plan

- [x] `npx prettier --check README.md` — clean (Markdown is prettier-checked in CI since #1294)
- [x] Rendered Markdown reviewed

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ DFX services can be integrated using a browser redirect to [app.dfx.swiss](https
 
 #### Iframe
 
-DFX services can be integrated by opening [app.dfx.swiss](https://app.dfx.swiss/) with the desired parameters (see [below](#query-parameters)) in an Iframe. See the [code example](#iframe-example) below. The `allow="clipboard-write"` attribute should be set on the Iframe, as copy buttons cannot use the browser clipboard from within a cross-origin Iframe without it.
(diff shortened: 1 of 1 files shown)


Danswar:
2 review passes. Corrected in the second: the initial rationale overclaimed ("copy buttons cannot use the browser clipboard without it" — the legacy fallback is not policy-gated and keeps working), now stated as the Chromium permissions-policy fact; and the example's self-closing `<iframe />` was invalid HTML, now properly closed.

Danswar:
Superseded by #1521.

PR #1300, opened 2026-08-10
d3628eb0 - Mock chargeback API dates as ISO strings in unit tests

TaprootFreak:
## Why

Follow-up to #1285 (see https://github.com/DFXswiss/services/pull/1285#issuecomment-5237772521): the unit tests mock the API payload with `Date` instances, but over JSON transport `requestedDate`/`date`/`chargebackDate` arrive as ISO strings. The rendering path handles both today, yet the tests would stay green through a regression that only breaks on string input.

## What

- Introduce a local transport-shaped `PendingChargebackPayload` type (dates as `string`, mirroring the e2e spec's fixtures) in both test files.
- Screen test: the fixture date constants and the sentinel `chargebackDate` are ISO strings now; `baseEntry()` builds the transport shape and casts once at the mock boundary.
- Hook test: the mocked payload is typed transport-shaped; no cast needed at the untyped `jest.fn()` boundary.
- Every successful mock payload is routed through a JSON round-trip helper (`asTransport`) at the mock boundary — mirroring what `response.json()` actually delivers: keys set to `undefined` are dropped, and the resolved value is a fresh clone, so reference aliasing cannot mask an in-place mutation. Rejection paths stay untouched. The hook test additionally asserts the raw ISO string on the result.
- No production code changes.

Both suites pass locally (2 suites, 9 tests).

### DIFF
--- a/src/__tests__/compliance-chargeback-list.screen.test.tsx
+++ b/src/__tests__/compliance-chargeback-list.screen.test.tsx
@@ -42,10 +42,16 @@ import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { ChargebackBlockReason, PendingChargebackEntry } from 'src/dto/chargeback.dto';
 import ComplianceChargebackListScreen from 'src/screens/compliance-chargeback-list.screen';
 
-const REQUESTED = new Date('2026-01-15T10:00:00.000Z');
-const DATE = new Date('2026-01-15T12:00:00.000Z');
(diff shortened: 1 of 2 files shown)


TaprootFreak:
Two review passes to zero findings. Pass 1 flagged two gaps in the new fixtures: the hook test compared the result against the very object reference it had fed into the mock, and some fixtures set keys to an explicit `undefined`, which a real JSON response never contains. Both are addressed by routing every successful mock payload through a JSON round-trip at the mock boundary and asserting the raw ISO string on the hook result.

PR #1305, opened 2026-08-10
1962fdea - Follow the CHF deposit cutover in the personal-IBAN frontend

TaprootFreak:
Follows the CHF deposit cutover (DFXswiss/api#4749): Bank Frick now serves CHF alongside EUR for personal-IBAN issuance and collection-account deposits. Without this change a CHF customer never sees the personal-IBAN presentation.

## Changes

- **`FRICK_COLLECTION_IBANS`** as the frontend's single guarded source (frozen map, own-key lookup); `FRICK_CURRENCIES` is derived from its keys and `isPersonalIbanApplicable` follows it, mirroring the backend's deposit-routing rule.
- **Per-currency collection IBANs**: `getFrickCollectionIban` serves the EUR and CHF Bank Frick collection accounts; the offer logic (`getOfferableCollectionIban`) returns the currency's IBAN or nothing — gate and value in one place, so the payment-info toggle can no longer disagree with its own precondition.
- **Stale EUR-only wording updated** in the mismatch hint, the currency error copy and the personal-IBAN promo gate (a CHF customer no longer gets steered towards EUR); de/fr/it translations renamed alongside their keys.
- **Tests**: EUR-only pins flipped onto the new contract with USD as the genuine non-Frick currency, new positive CHF wiring tests, full branch coverage on the reworked util.
- **Handbook**: new Playwright baselines for the CHF collection-IBAN toggle, the CHF buy page (no mismatch, no promo), the USD mismatch hint (updated copy), the USD promo state, and a first `buy/info` spec with its `metadata.json` entry. The CHF fixture IBAN is check-digit valid (mod-97).

## Notes for review

- The mismatch hint and the promo are mutually exclusive render branches (a recognized selector always suppresses the promo); the baselines therefore prove them in separate scenarios.
- Existing baselines are untouched: the wallet-2 page shots capture the e-mail gate state, which never renders the promo.
- fr/it copy for the two renamed keys is a direct translation following the original sentence pattern — wording review welcome.


TaprootFreak:
3 review passes until clean (1 pre-PR, 2 on the PR). Fixed along the way: the collection-IBAN offer now returns the currency's IBAN from a single guarded map (own-key lookup, currency list derived from it), the CHF quote is proven to request the Frick provider end to end, explicit negative assertions guard against the EUR collection IBAN or the promo/hint leaking into the CHF states, the promo baseline captures the promo block itself (the full-page shot capped out before it), and every fixture IBAN in the diff is check-digit valid.

PR #1307, opened 2026-08-10
137676bf - Guard every /compliance path, including ones no screen claims

### LINKED ISSUE #1306: Compliance guard does not redirect a plain User away from /compliance/pending-chargebacks
## What happens

A logged-in user with the plain `User` role can open `/compliance/pending-chargebacks` and stays there. Every other compliance route redirects that role away within milliseconds.

Observed in a full run of the e2e-stack harness: the poll below waits the full 15s and the path never changes.

```
Error: User role must be redirected away from /compliance/pending-chargebacks
Expected: not "/compliance/pending-chargebacks"
Call Log:
- Timeout 15000ms exceeded while waiting on the predicate
```

The same assertion passes for all 23 other compliance routes in the same test run, so this is specific to this route and not a harness or session issue.

## Why it is surprising

`ComplianceChargebackListScreen` calls `useComplianceGuard()` as its first statement, exactly like its siblings:

```tsx
export default function ComplianceChargebackListScreen(): JSX.Element {
  useComplianceGuard();
  useLayoutOptions({ title: 'Pending Chargebacks', backButton: true, noMaxWidth: true });
```

The guard is therefore present but does not take effect here. The route definition in `App.tsx` is structurally identical to the neighbouring compliance routes. Root cause not established — that is what this issue is for.

## Impact

Unclear from the outside, and worth checking rather than assuming: whether the screen merely fails to redirect, or whether it also renders chargeback data to an unauthorised role. `getPendingChargebacks()` is called from a `useEffect` gated only on `isLoggedIn`, not on the role.

## Where it comes from

The screen was added in #1285. Not introduced by the e2e harness — the harness only made it visible.

## Current state in the test suite

Recorded as a known product bug in `e2e-stack/specs/compliance.spec.ts` via `test.fail()`, rather than dropped from the coverage list, so the gap stays visible. When the guard is fixed, that test starts failing and is the reminder to turn it back into a regular assertion and to move the route back into `ALL_COMPLIANCE_PATHS`.

TaprootFreak:
Closes #1306.

## The defect

A URL below `/compliance` that matches no route — a typo, a screen not deployed yet, a screen removed —
makes React Router raise a 404 `ErrorResponse` and render the root `errorElement` (`ErrorScreen`). That
screen runs **outside every guard**, and an error boundary does not change the URL. An unauthorised role
therefore sits on a `/compliance/…` address and is never sent away.

That is what #1306 observed. The report named `/compliance/pending-chargebacks`, and the guard on that
screen turned out to be intact — the run had been made against a build without the route
(`feat/e2e-stack` branches off #1282, the screen landed with #1285, and the harness builds the bundle from
its own working tree). But the hole underneath is real and independent of that one route: **the role check
for the compliance area only exists where a screen exists.**

## The fix

A guarded catch-all as the last compliance route:

- `src/screens/compliance-not-found.screen.tsx` — calls `useComplianceGuard()` like every other compliance
  screen, then states plainly that the page does not exist and shows the requested path.
- `src/App.tsx` — `path: 'compliance/*'`, placed after all compliance routes. React Router ranks static
  segments above a splat, so it only ever matches what no screen claims.
- `src/__tests__/compliance-not-found.screen.test.tsx` — pins both halves: the guard call (with an empty
  argument list) and the rendered message plus requested path.

The role check for `/compliance` no longer depends on whether a given screen exists, and an authorised
user gets "this compliance page does not exist" instead of "something went wrong".

Behaviour outside `/compliance` is untouched: an unknown path elsewhere still renders the generic error
screen exactly as before.

## Second change: the guard assertion the screen tests were missing

`src/__tests__/compliance-chargeback-list.screen.test.tsx` mocked `useComplianceGuard` away and never
asserted the call — deleting the guard from the screen left all eight tests green. It now asserts the call
**with an empty argument list**, because `useComplianceGuard(redirectPath = '/', isActive = true)`
navigates only while `isActive`: `useComplianceGuard(undefined, false)` would keep the call and still leave
the guard inert.

## Evidence

Measured against a running dev server, role taken from a JWT (the frontend decodes it client-side, so the
guard decision involves no backend), polling `location.pathname`:

| Path | Before | After |
| --- | --- | --- |
| `/compliance/definitiv-nicht-existent` (invented) | stays on the path, "something went wrong" | **redirected to `/`** |
| `/compliance/pending-chargebacks` | redirected to `/` | redirected to `/` |
| `/compliance/recalls` | redirected to `/` | redirected to `/` |
| `/voellig-unbekannt-ausserhalb` (outside `/compliance`) | stays, generic error | stays, generic error (unchanged) |

The splat claims nothing that belongs to a real screen — verified by which lazy chunk each URL loads:

```
/compliance/kyc-stats                → compliance-kyc-stats
/compliance/recalls                  → compliance-recall-list
/compliance/pending-chargebacks      → compliance-chargeback-list
/compliance/call-queues/foo/1        → compliance-call-queue-detail
/compliance/mros/create              → compliance-mros-create
/compliance/definitiv-nicht-existent → compliance-not-found
```

Mutation checks on the guard assertion, each restored afterwards: removing `useComplianceGuard();` and
changing it to `useComplianceGuard(undefined, false);` each fail that one test and only that test.

Full suite: 81 suites, 950 tests, all green. ESLint clean on both new files.

## Coverage

- `src/screens/compliance-not-found.screen.tsx` — **100 % statements / 100 % branches / 100 % functions /
  100 % lines**.
- `src/screens/compliance-chargeback-list.screen.tsx` (untouched, measured with the new assertion in
  place) — **100 % / 100 % / 100 % / 100 %**.
- `src/App.tsx` is touched by two insertions (a lazy import and a route entry). It is the routing table,
  has no test of its own in this repository, and is not brought to 100 % here — a **declared deviation**
  from the coverage rule in CONTRIBUTING.md, and the reviewer's call. The two added lines are exercised in
  the browser evidence above.

## Note for #1288

Once that branch is rebased onto a `develop` containing #1285, `/compliance/pending-chargebacks` belongs in
`ALL_COMPLIANCE_PATHS` as a normal assertion and both `test.fail()` markers can go: the redirect assertion
now passes through this catch-all even while a screen is missing, and the mount assertion passes once the
screen is present.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -76,6 +76,7 @@ const ComplianceReviewScreen = lazy(() => import('./screens/compliance-review.sc
 const ComplianceCallQueuesScreen = lazy(() => import('./screens/compliance-call-queues.screen'));
 const ComplianceCallQueueScreen = lazy(() => import('./screens/compliance-call-queue.screen'));
 const ComplianceCallQueueDetailScreen = lazy(() => import('./screens/compliance-call-queue-detail.screen'));
+const ComplianceNotFoundScreen = lazy(() => import('./screens/compliance-not-found.screen'));
 const SupportDashboardOverviewScreen = lazy(() => import('./screens/support-dashboard-overview.screen'));
 const SupportDashboardScreen = lazy(() => import('./screens/support-dashboard.screen'));
 const SupportDashboardIssueScreen = lazy(() => import('./screens/support-dashboard-issue.screen'));
@@ -461,6 +462,12 @@ export const Routes = [
       {
         path: 'compliance/call-queues/:queue/:userDataId',
         element: withSuspense(<ComplianceCallQueueDetailScreen />),
+      },
+      // Must stay last among the compliance routes: React Router ranks static segments above a splat,
+      // so this only matches paths none of the screens above claim.
+      {
+        path: 'compliance/*',
+        element: withSuspense(<ComplianceNotFoundScreen />),
       },
       {
         path: 'sitemap',
--- a/src/__tests__/compliance-chargeback-list.screen.test.tsx
+++ b/src/__tests__/compliance-chargeback-list.screen.test.tsx
@@ -5,6 +5,7 @@
 let mockIsLoggedIn = true;
 const mockGetPendingChargebacks = jest.fn();
 const mockNavigate = jest.fn();
+const mockUseComplianceGuard = jest.fn();
 
 jest.mock('@dfx.swiss/react', () => ({
   useSessionContext: () => ({ isLoggedIn: mockIsLoggedIn }),
@@ -23,7 +24,7 @@ jest.mock('src/components/error-hint', () => ({
 }));
 
 jest.mock('src/hooks/guard.hook', () => ({
-  useComplianceGuard: () => undefined,
+  useComplianceGuard: () => mockUseComplianceGuard(),
 }));
 
 jest.mock('src/hooks/layout-config.hook', () => ({
@@ -102,6 +103,15 @@ describe('ComplianceChargebackListScreen', () => {
     jest.clearAllMocks();
     mockIsLoggedIn = true;
     mockGetPendingChargebacks.mockResolvedValue(asTransport([]));
+  });
+
+  // The role guard is all that stands between an unauthorised role and this screen; it is mocked out
+  // in every other test here. Without this assertion the screen could lose its guard call and the
+  // whole suite would stay green — see issue #1306, where exactly that was suspected.
+  it('calls the compliance guard on render', () => {
+    render(<ComplianceChargebackListScreen />);
+
+    expect(mockUseComplianceGuard).toHaveBeenCalled();
   });
 
   it('does not fetch when not logged in', () => {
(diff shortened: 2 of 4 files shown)


TaprootFreak:
**EN:** Three review passes, the last two triggered by findings this branch fixed — a guard assertion that ignored its arguments, then the catch-all route that closes the underlying hole.
**DE:** Drei Review-Durchläufe, die letzten beiden ausgelöst durch Befunde, die dieser Branch behoben hat — eine Guard-Assertion, die ihre Argumente ignorierte, und danach die Catch-all-Route, die die eigentliche Lücke schliesst.

<details>
<summary>Details</summary>

**Pass 1** — both dimensions clean on the first commit, which added the missing guard assertion to the chargeback screen test.

**Pass 2** — the logic dimension pointed out that the guard mock discarded its arguments, so `useComplianceGuard(undefined, false)` would have kept the call in place, left the guard inert, and still passed. The mock now forwards its arguments and the test asserts an empty argument list. Both mutations (call removed, call made inert) fail that test and only that test. Both dimensions clean afterwards.

**Pass 3** — covers the product fix: a guarded catch-all for `/compliance/*`, so an unknown compliance path can no longer park an unauthorised role on a `/compliance` URL. Both dimensions clean, including the check that the splat claims nothing belonging to a real screen (verified independently against the route table and, in the browser, against the lazy chunk each URL loads).

</details>

PR #1308, opened 2026-08-10
406da72e - Document the test architecture, require a reality declaration, and fix the E2E full-run flag

TaprootFreak:
Documents the test layers this repository owns with measured numbers, links to the canonical cross-repository description, and makes a reality declaration mandatory whenever a test setup replaces something real with a fake.

## Why

The unit suite, the full-stack harness and the visual-regression suite answer three different questions, and what each one fakes was described across several files. The practical consequence: a full-stack run can be green while no layer verifies a payment end to end, because every process-gated background job is switched off during that run. That should be readable in one place instead of measured.

## What this adds

`docs/test-architecture.md` for the three layers this repository owns, stating for each what it proves, what it deliberately does not prove, and whether it runs in CI. It also states plainly that the processing chain behind the API — incoming transfers, AML, purchase calculation, liquidity, payout, ledger booking — is not testable from here and belongs to the integration layer in `DFXswiss/api`.

Current state is separated from target architecture, and every number carries the commit it was measured on plus the command that reproduces it. Measured on `develop` at `4e9544a9`:

| Metric     | Coverage | Absolute     |
| ---------- | -------- | ------------ |
| Statements | 18.15 %  | 2 578/14 198 |
| Branches   | 17.31 %  | 2 054/11 861 |
| Functions  | 14.44 %  | 668/4 626    |
| Lines      | 18.63 %  | 2 375/12 742 |

That number is what the existing coverage rule has to be read against: every file a pull request touches must reach 100 % on all four metrics, and CI does not enforce it — it is a review gate. At 18 %, touching a long-neglected file makes its whole coverage an obligation, which is worth planning for rather than discovering in review.

## Written against a moving target

The document was first written while #1288 was still open, so every mention of the full-stack layer was a forward reference. #1288 has since been merged, this branch was rebased onto it, and the document now describes that layer as present, with the unit figures re-measured on the merged develop: `e2e-stack/`, its registry and the route gate are on `develop`. The figures for the suite keep the commit they were measured on — the head of that pull request — with a note to re-measure after changes.

Both references between the two files are plain paths rather than markdown links, on purpose: the handbook build renders every markdown file to HTML, and its integrity check then requires every relative reference beginning with `docs/`, `assets/` or `screenshots/` to exist in that output — which a reference to a `.md` file never does. That check turned the container-smoke job red once during this work. References outside those three prefixes are unaffected, which is why the existing link to `e2e-stack/README.md` is fine.

## The reality declaration

Whenever a pull request introduces, removes or changes a fake — a faked external provider, a disabled cron job, a schema built without the migration chain, state written directly with SQL, a placeholder value that looks real, a suppressed side effect, or a seed correction that bends reality — the declaration changes in the same pull request, and each entry says in one plain sentence what a green run does **not** prove. The entry is written before the fake is built; reversed, it becomes documentation written from memory, with omissions.

The full definition lives in `DFXswiss/api` so the two repositories cannot drift apart on it. The measurement principle follows the route gate already used here: verify what the run actually did, not what someone declared.

## Also fixes the full-run flag

`e2e-stack/scripts/run.sh` set `E2E_FULL_RUN=1` on every invocation while forwarding `"$@"` — the very way a filter reaches Playwright. So `npm run e2e:stack -- --grep x` declared a full run and executed a subset, and the route-coverage gate then checked its claim list against a navigation recording that could only be partial, reporting routes as never opened because their specs had been filtered out. The comment directly above that line already stated the rule the line broke.

The flag is now set only when no arguments were given, and a filtered run says so in its log. CI is unaffected: `e2e-stack.yml` does not use `run.sh` and sets the variable itself for its own unfiltered invocation. The documentation names both sources of the flag, so rule and implementation can be checked against each other.

This was found while reviewing the documentation of that gate. It is fixed here rather than deferred, because a defect in code a pull request touches gets fixed in that pull request.

## Coverage

None of the three files this pull request touches is instrumented by Jest: `collectCoverageFrom` in `package.json` covers `src/**/*.{ts,tsx,js,jsx}` minus declarations, while the changes are `CONTRIBUTING.md`, `docs/test-architecture.md` and `e2e-stack/scripts/run.sh`. There are therefore no per-file coverage figures to state, and no under-fulfilment of the 100 % rule to declare.

## Scope

Documentation only — no code, no workflow and no test changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,9 +42,12 @@ reason to leave the bug unreported.
 measured numbers for the current state, and points at the canonical
 cross-repository description in `DFXswiss/api`. Read it before adding a test
 layer, moving a test between layers, or extending the full-stack harness. (Given
-as a path rather than a link on purpose: the handbook build renders every
-markdown file to HTML and its integrity check rejects a link that resolves to a
-`.md` file, which is why no markdown file in this repository links to another.)
+as a path rather than a link on purpose: the handbook build renders every markdown
+file to HTML, and its integrity check then requires every relative reference
+beginning with `docs/`, `assets/` or `screenshots/` to exist in that output — which
+a reference to a `.md` file never does, since only the rendered `.html` is there.
+References outside those three prefixes are unaffected, which is why the link to
+`e2e-stack/README.md` further down is fine.)
 
 Two obligations follow from it for every pull request:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,11 +38,13 @@ reason to leave the bug unreported.
 
 ### Test architecture
 
-[docs/test-architecture.md](docs/test-architecture.md) describes the test layers
-this repository owns, with measured numbers for the current state, and links to
-the canonical cross-repository description in `DFXswiss/api`. Read it before
-adding a test layer, moving a test between layers, or extending the full-stack
-harness.
+`docs/test-architecture.md` describes the test layers this repository owns, with
+measured numbers for the current state, and points at the canonical
+cross-repository description in `DFXswiss/api`. Read it before adding a test
+layer, moving a test between layers, or extending the full-stack harness. (Given
+as a path rather than a link on purpose: the handbook build renders every
+markdown file to HTML and its integrity check rejects a link that resolves to a
+`.md` file, which is why no markdown file in this repository links to another.)
 
 Two obligations follow from it for every pull request:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,28 @@ reason to leave the bug unreported.
 
 ## Testing
 
+### Test architecture
+
+[docs/test-architecture.md](docs/test-architecture.md) describes the test layers
+this repository owns, with measured numbers for the current state, and links to
+the canonical cross-repository description in `DFXswiss/api`. Read it before
+adding a test layer, moving a test between layers, or extending the full-stack
+harness.
+
+Two obligations follow from it for every pull request:
+
+- **A failure mode is tested at the lowest layer that can express it.** A case
+  reachable against a real database does not belong in a browser test, and the
+  processing chain behind the API is not testable from this repository at all.
+- **Reality declaration.** Whenever a pull request introduces, removes or changes
+  a fake — a faked external provider, a disabled cron job, a schema built without
+  the migration chain, state written directly with SQL, a placeholder value that
+  looks real, a suppressed side effect, or a seed correction that bends reality —
+  the declaration changes in the same pull request, and each entry says in one
+  plain sentence what a green run does **not** prove. Write the entry before
+  building the fake. A pull request that adds a fake without its declaration is
+  incomplete regardless of whether CI is green.
+
 ### Unit tests
 
 ```
--- a/docs/test-architecture.md
+++ b/docs/test-architecture.md
@@ -52,13 +52,12 @@ the merge target had gained a route the registry did not claim yet. Re-measure a
 suite; the number of tests is not pinned anywhere.
 
 The harness runs the following for real: Postgres, the API, this frontend, a browser. It fakes every
-external provider, on two levels — the API mocks its own outbound calls, and the Docker network it sits
-on has no route to the internet at all. The second level is the load-bearing one, and
-`e2e-stack/env/api.env` records why: the `loc` mock covers only calls made through the API's central
-HTTP wrapper, so an integration reaching out through a vendor SDK, `graphql-request`, an ethers provider
-or bare `fetch` goes straight past it. What guarantees that no external system is contacted is therefore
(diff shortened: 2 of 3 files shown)


TaprootFreak:
**EN:** Interim note from an earlier stage of this pull request — the final review summary is the last comment below.

**DE:** Zwischenstand aus einer früheren Phase dieses Pull Requests — die abschliessende Review-Zusammenfassung steht im letzten Kommentar unten.

TaprootFreak:
**EN:** Nine review passes were needed to reach zero blocking findings; the pull request also grew a code fix, because a defect found in review gets fixed rather than filed.

**DE:** Neun Review-Durchläufe waren nötig, um null blockierende Funde zu erreichen; der Pull Request enthält zusätzlich einen Code-Fix, weil ein im Review gefundener Defekt behoben und nicht erfasst wird.

<details>
<summary>Details</summary>

A document that makes "current state and target are kept apart" a rule has to survive being held to it. It did not, repeatedly.

## What the passes found

**Pass 1 — five blocking findings, all the same class.** The layer table listed `e2e-stack/` as a present location running in CI, two references pointed at paths that did not exist on the branch, and the route-coverage section described the gate in the present tense. All of it was a forward reference to the then-unmerged harness, none of it marked as such.

**Pass 2 — clean.** At that point the pull request looked finished; it was set ready and had to be pulled back, see below.

**Pass 3 — one blocking finding.** The note explaining why the document is referenced by path claimed the handbook check rejects any link resolving to a `.md` file, and that no markdown file here links to another. Both were too broad: the check only inspects references beginning with `docs/`, `assets/` or `screenshots/`, and `CONTRIBUTING.md` has linked `e2e-stack/README.md` since #1288.

**Pass 4 — one blocking finding.** After #1288 merged, the whole document was rewritten from forward reference to present tense. The rewrite left behind a claim that around twenty integrations bypass the API's central HTTP wrapper — a number this repository cannot check, and unlike its neighbours it carried no attribution.

**Pass 5 — one blocking finding, again introduced by the previous fix.** The replacement credited `e2e-stack/env/api.env` with naming vendor SDKs, `graphql-request`, ethers providers and bare `fetch`. That file names none of them.

**Pass 6 — three precision points, all taken.** The network isolation is not a fake, so describing it as one of two ways providers are "faked" was wrong; `one database and one API instance` now cites `e2e-stack/compose.yml`, which declares exactly one of each; and the effect of `DISABLED_PROCESSES=*` inside the API is attributed rather than asserted.

**Pass 7 — clean, in both dimensions.**

## Passes 8 and 9 — the code fix, and a gap inside it

Reviewing the documentation of the route gate surfaced a defect in the gate's own launcher, `e2e-stack/scripts/run.sh`. It set `E2E_FULL_RUN=1` on every invocation while forwarding `"$@"` — the very way a filter reaches Playwright — so `npm run e2e:stack -- --grep x` declared a full run and executed a subset. The comment directly above that line already stated the rule the line broke. The gate would then compare its claim list against a navigation recording that could only be partial and report routes as never opened because their specs were filtered out.

Pass 8 found that the first fix was incomplete: setting the flag only in the unfiltered branch is not the same as guaranteeing it is unset in the filtered one, because `e2e-stack/compose.tests.yml` forwards `${E2E_FULL_RUN:-}` and an inherited value would survive. The filtered branch now clears it explicitly. Pass 9 confirmed the whole thing: the flag arrives empty on a filtered run, `route-coverage.spec.ts` compares strictly against `'1'` so empty behaves as unset, and the unfiltered path is byte-identical to before.

No call path in this repository set the variable ahead of `run.sh`, so this closed a latent gap rather than an active break — the CI workflow does not use `run.sh` and sets the variable itself.

## What was corrected in the rules themselves

Two findings were not about facts but about the rules this document sets. The closing rule now applies to **measured** numbers: as written before, it could be read to demand a measurement stamp for phrases like "that any two parts fit together", which one review pass duly flagged. And the reality-declaration section no longer quotes a category and field count it cannot verify from here — that taxonomy belongs to the companion document.

## What was rejected

A final pass marked five statements as blocking because they describe the other repository without a "not verifiable from here" clause — that the canonical description lives in `DFXswiss/api`, that migrations are covered there, that it checks this repository out to obtain the harness. Those were rejected after verification: all four are accurate, they are plain location statements rather than unknown quantities, they sit inside the cross-repository framing the document opens with, and this document's own standard asks for a stamp on measured numbers, not on pointers. Adding the clause to each would cost readability and protect nobody.

## Related

Issue #1313 described that defect and is superseded by the fix in this pull request; it will be closed on merge.

## Note on the earlier ready state

This pull request was briefly set ready and pulled back to draft. The handbook check had been SKIPPED at the time of that decision and ran afterwards, failing on the markdown link described under pass 3. The link is gone; the check has been green since.

## Scope

Documentation plus one fix to `e2e-stack/scripts/run.sh`; no workflow and no test changes. `run.sh` is outside the coverage globs in `package.json`, so the per-file coverage rule does not apply to it. Unit figures measured on `develop` at `4e9544a9`; the suite figures keep the commit they were taken on.

</details>

Issue #1313, opened 2026-08-10
e2e-stack: run.sh declares a full run even when it forwards filter arguments


TaprootFreak:
`e2e-stack/scripts/run.sh` sets `E2E_FULL_RUN=1` on every invocation while also forwarding any arguments it was given:

```sh
# Declares that every spec is in scope, which is what lets the coverage gate check that each route
# was actually opened rather than merely claimed. A filtered run must not set this.
E2E_FULL_RUN=1 compose run --rm tests "$@"
```

The comment states the rule correctly — a filtered run must not set the flag — and the line below it breaks it, because `"$@"` is exactly how a filter reaches Playwright. So `npm run e2e:stack -- --grep @something` declares a full run while executing a subset.

What that costs: `e2e-stack/specs/route-coverage.spec.ts:306` only performs the "was this route actually opened" half of the gate when the flag is `1`, and prints an explicit notice at line 322 when it is not. With the flag wrongly set on a filtered run, the gate compares the claim list against a navigation recording that was never going to be complete, so it reports routes as never opened purely because their specs were filtered out.

Two ways to fix it:

1. Only set the flag when no arguments were passed — `if [ $# -eq 0 ]; then E2E_FULL_RUN=1; fi` — which matches the comment exactly.
2. Refuse the combination: if arguments are present, either drop the flag or fail with a message saying a filtered run cannot satisfy the route gate.

Found while reviewing #1308, which documents this gate. Not caused by that pull request and not blocking it; reported here because a defect in code a pull request touches is reported regardless of who introduced it.

TaprootFreak:
**EN:** Fixed in #1308 rather than left for later — `run.sh` now sets the full-run flag only for an unfiltered run and clears it otherwise.

**DE:** In #1308 behoben statt vertagt — `run.sh` setzt das Full-Run-Flag nur noch für einen ungefilterten Lauf und löscht es andernfalls.

<details>
<summary>Details</summary>

`e2e-stack/scripts/run.sh` now sets `E2E_FULL_RUN=1` only when it was given no arguments, and clears it explicitly on a filtered run. Clearing matters rather than merely not setting: `e2e-stack/compose.tests.yml` forwards `${E2E_FULL_RUN:-}`, so a value inherited from the caller's environment would otherwise survive and make a filtered run declare itself complete.

[...]

Issue #1312, opened 2026-08-10
e2e-stack: route coverage is not claim-bound, and factory id hardening is incomplete


TaprootFreak:
Raised in the review of #1288 and deferred so the PR can merge. Both were classified as medium, below the gate-integrity blockers.

### Route coverage is a global path set, not claim-bound

Navigation recording writes pathnames only, so the gate asks "was this route opened by *anyone*?" rather than "by the suite named in `claim.spec`?". That under-delivers what CONTRIBUTING promises — a registry entry pointing at a file that never visits the route should fail, but today another suite incidentally visiting the path satisfies it.

**Fix:** record `{ pathname, specFile }` (e.g. via `testInfo.file`) and match each claim against navigations from its own `claim.spec`.

### Factory 2xx-without-id hardening is incomplete

`createLimitRequest` throws when no id comes back, but `createBankAccount`, `createBuy`, `createSell` and `createSwap` still trust `res.id` / `res.routeId`, and `track()` silently drops non-numeric ids.

**Fix:** a shared `requirePositiveId(...)` applied after every create response.

TaprootFreak:
**EN:** Both points are closed by #1314, verified on `develop`.
**DE:** Beide Punkte sind mit #1314 geschlossen, auf `develop` verifiziert.

<details>
<summary>Details</summary>

- **Claim-bound route coverage** — navigations record the spec file that drove them, and each claim
  is matched only against navigations from its own `claim.spec`. Both sides derive the spec-file
  form through the same function, so a suite moving into a subdirectory cannot make its own claims
  look unvisited. The two failure modes are reported apart: never opened, and opened by the wrong
  suite.
- **Factory id hardening** — a shared `requireId()` guards every create response and `track()`
  itself, demanding a genuine positive integer. It also surfaced a real defect on the way:
  `createLimitRequest` returned a row resolved from the response body without ever registering it
  for cleanup.

Remaining, filed separately: #1315 (a splat claim is credited for any navigation below it).

</details>

Issue #1311, opened 2026-08-10
e2e-stack: three gate-integrity holes (forbidOnly, .env in build context, swallowed cleanup failures)


TaprootFreak:
Raised in the review of #1288 and deferred so the PR can merge. All three let the E2E gate report green without actually having checked — they are about the trustworthiness of the gate itself, not about the product under test.

### 1. `forbidOnly` is missing (HIGH)

`e2e-stack/playwright.config.ts` sets `workers: 1` and `retries: 0` but no `forbidOnly`. A committed `test.only` would skip the bulk of the suite **and** the `coverage-gate` project, and CI would still be green.

**Fix:** `forbidOnly: true`, or at minimum `forbidOnly: !!process.env.CI`.

### 2. Root `.dockerignore` does not exclude `.env` / `.env.generated` (HIGH)

The frontend build context is the repo root (`compose.yml`: `context: ..`) and the Dockerfiles do `COPY . .`. The root `.dockerignore` excludes `node_modules`, `api-repo` and build artefacts — but not `.env`, `**/.env`, `**/.env.*`, and specifically not `e2e-stack/.env.generated`, which `up.sh` writes before `compose build frontend`.

Being gitignored does not help here: the files still travel to the Docker daemon and into builder layers, and CRA can bake `REACT_APP_*` values into the bundle.

**Fix:** add `.env`, `**/.env`, `**/.env.*` and `e2e-stack/.env.generated` to the root `.dockerignore`, keeping sample files via a negation if needed.

### 3. `cleanupCreatedData` swallows delete failures (HIGH)

`e2e-stack/specs/fixtures/factories.ts` (~1657–1697) collects delete errors and only `console.warn`s them; every `afterAll` ignores the return value. Leftover rows in the shared database can make later specs false-green or false-red without ever failing the file that caused it. The registry is also cleared **before** the deletes (`created.length = 0`), so failed rows are never retried, and the comment there explicitly defends the best-effort behaviour — which is the opposite of what a shared-DB gate needs, and contradicts the repository rule against silent fallbacks.

**Fix:** keep failed references registered until they are actually deleted, and throw (e.g. an `AggregateError`) when any delete failed, so the originating spec goes red. Successful cleanup must stay silent, otherwise the whole suite turns red.

TaprootFreak:
**EN:** All three gate-integrity holes are closed by #1314, verified on `develop`.
[...]

PR #1314, opened 2026-08-10
4fd00f34 - Close the E2E gate-integrity holes and bind route coverage to its claim

TaprootFreak:
Closes the findings that were filed rather than fixed while #1288 was merged. All of them are about the trustworthiness of the gate itself: each one lets the harness report green without having checked.

## Gate integrity (#1311, #1313)

**`forbidOnly`** — the config pinned `workers: 1` and `retries: 0` but never forbade `test.only`. A committed one would have skipped the bulk of the suite *and* the `coverage-gate` project while CI stayed green.

**`.env` in the Docker build context** — the frontend build context is the repository root and the Dockerfiles `COPY . .`, but the root `.dockerignore` excluded only `node_modules`, `api-repo` and build artefacts. Gitignore does not protect a build context: an untracked `.env` still travels to the daemon and into builder layers, and CRA bakes `REACT_APP_*` values into the bundle. `up.sh` writes `e2e-stack/.env.generated` before `compose build frontend`, so this was concrete, not theoretical.

**Swallowed cleanup failures** — `cleanupCreatedData` cleared the registry *before* the deletes and only warned on failure, while every `afterAll` discards its return value. On a shared database a leftover row is state the next spec file inherits. It now throws, and failed references are re-registered in their original order so the next call retries them. Successful cleanup stays silent. The harness `tsconfig.json` moves to the ES2021 lib for the `AggregateError` type.

**`E2E_FULL_RUN` on filtered runs** — `run.sh` set the flag unconditionally while forwarding `"$@"`, which is how `--grep` reaches Playwright. A filtered run therefore declared itself complete. It now sets the flag only for a genuinely unfiltered run, decided by an **allowlist**: only flags known not to change *which* specs run leave a run complete — `--workers`, `--reporter`, `--timeout`, `--retries` and their kind, in both the `=` and the separate-value form, whose value is skipped rather than mistaken for a spec path. Everything else counts as a filter, including flags Playwright may add after this was written.

The list started as the opposite and was wrong twice: `--project`, `--shard`, `--last-failed` and `--only-changed` were each missed in turn, and every omission let a partial run declare itself complete. An allowlist fails the other way — an unrecognised flag costs the strict navigation check on a local run, which is the harmless direction. `--config` and `-c` are deliberately off it too: another config can drop the coverage-gate project entirely, so a flag that can replace the whole selection cannot be one that promises the selection is complete.

`develop` solved the same problem in parallel through #1308. Two details of that solution are better and are kept here: the variable is passed empty rather than merely left unset, because `compose.tests.yml` forwards `${E2E_FULL_RUN:-}` and a value exported by the calling shell would otherwise survive into the container; and a filtered run now says so on stdout.

## Cleaning up along the foreign keys

The throw immediately showed that cleanup had never worked completely — six spec files failed on foreign key violations from rows the application itself wrote and no test registered: `support_message` under `support_issue`, `buy_crypto` under both `bank_tx` and `buy`, `bank_tx` under `transaction`. The old helper covered two parents and descended one level.

Cleanup now reads the foreign keys once and descends into whatever references a row before deleting it, children first. Every statement still names a concrete id or foreign key value, so it can no more touch another account's rows or seed data than before. Cycles and diamonds terminate through a result map that returns what actually happened on a repeat visit rather than assuming success.

Whether a child table is a leaf is decided by whether anything references it — not by the shape of its primary key. A table with no primary key can still be referenced through a UNIQUE column; a composite-key join table nothing points at, like `mros_transactions_transaction`, has no descendants to lose and is deleted through its foreign key column. A table that is referenced but not addressable by `id` is reported through the same error path as an unresolvable foreign key, and only when a row for that parent actually exists.

## Claim-bound route coverage (#1312, #1309)

Navigation recording stored pathnames only, so the gate asked whether a route had been opened by *anyone* rather than by the suite named in `claim.spec`. Navigations now carry the spec file that drove them, and the gate matches each claim against navigations from its own file. The two failure modes are reported apart — never opened, and opened by the wrong suite — and the second names the files that did open it. Both sides derive the spec-file form through the same function, relative to `specs/`, so a suite moving into a subdirectory cannot silently make its own claims look unvisited.

## Factory id hardening (#1312, #1310)

`createLimitRequest` already refused a 2xx without an id; `createBankAccount`, `createBuy`, `createSell` and `createSwap` trusted the response body, and `track()` dropped a non-numeric id silently. A shared `requireId()` now guards every create response and `track()` itself, demanding a genuine positive integer — a stringified or fractional id is exactly the broken contract it exists to catch. `createSupportIssue` no longer tolerates a 2xx whose row it cannot find, and the one place where an absent id is legitimate says so with an explicit check rather than relying on truthiness.

One real defect surfaced on the way: `createLimitRequest` tracked `issueRow.limitRequestId` while returning `issueRow.limitRequestId ?? res.limitRequest?.id`, so a row resolved from the response body was returned but never registered for cleanup.

## Not in this pull request

Two pre-existing findings are filed separately: route coverage credits a splat claim for any navigation below it (#1315), and the module-load wallet-counter seeding has no rejection handler (#1316).

## Verification

`npx tsc --noEmit` under `e2e-stack/tsconfig.json` passes on Node 20, the version CI pins. The claim-bound gate and the recursive cleanup are only really exercised by a full stack run, which is what the `Full-stack E2E` check on this pull request does.

## What the review rounds changed

Four rounds of independent review, and the two most serious findings were in the cleanup this pull
request had just rewritten.

**A composite foreign key was paired by name, not by position.** The catalogue joined
`key_column_usage` to `constraint_column_usage` on the constraint name alone. For a key over several
columns that is a cross product: `(a, b) REFERENCES parent (x, y)` came back as four pairs, two of
which do not exist — and cleanup then used each pair on its own to find rows by that column, so it
could have found, and deleted, rows belonging to a different parent. It reads `pg_constraint` now
and pairs `conkey` to `confkey` by ordinality. Proven against a throwaway Postgres carrying one
two-column and one single-column key: the old query returns `a→x, a→y, b→x, b→y`, the new one
returns `a→x` and `b→y`, and the single-column case is identical either way.

**The first fix for self-referencing tables was itself the same mistake.** It refused to bulk-delete
as soon as the *schema* declared a foreign key onto itself — so a table that merely *can* reference
itself, holding one row whose self-reference is null, became unresolvable, and the parent delete
then failed on the foreign key. Deciding from the schema rather than from the rows present is
exactly what this pull request set out to remove. The gate now probes whether a row actually
references one of the rows about to be deleted, scoped to the one parent value. Rows in the same
batch count as referencers on purpose: a multi-row `DELETE` gives no order to rely on.

Three regression tests were added with it, each against a schema it creates and drops: composite
keys keep their ordinality, a table with an *unused* self key is cleaned up, and a table with a real
self reference is refused with its rows still present. A key spanning several columns is probed one
pair at a time, which can only refuse a delete that would have been safe, never allow one that is
not — the catalogue carries no constraint identity to group pairs by. That limit is stated in the
code and in `test-data.md`.

Also from the rounds: the claim-binding test now runs against a pure function with synthetic data,
so it can state the case the real registry cannot — a route opened, but by a different spec file
than the one claiming it. And two comments that described the opposite of the code were corrected.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,3 +14,13 @@ playwright-report
 # image from this directory. Without this line the whole of it travels to the Docker daemon as
 # build context on every image build.
 api-repo
+# Gitignore does not protect the Docker build context: context is whatever is on disk, not what
+# Git tracks. An untracked or gitignored .env still reaches the daemon and lands in builder
+# layers — and CRA (frontend / frontend-widget) bakes any REACT_APP_* values it finds in a .env
+# file into the JS bundle at build time. e2e-stack/scripts/up.sh also writes
+# e2e-stack/.env.generated with build-time-resolved values before `compose build frontend`; that
+# file must stay out of the context too.
+.env
+**/.env
+**/.env.*
+e2e-stack/.env.generated
--- a/e2e-stack/docs/test-data.md
+++ b/e2e-stack/docs/test-data.md
@@ -218,6 +218,8 @@ the "level 50 but no limit granted" case instead.
 
 Deletes every row tracked during this process (`{ table, id }`) in reverse creation order, descending into rows that reference them by foreign key first — including rows the application itself wrote and no test registered. Returns `{ deleted, errors }` on success and stays silent; if any delete fails, the failed references are re-registered for the next call and it throws an `AggregateError`, so the spec file that caused the leftover goes red rather than an unrelated one later against the shared database.
 
+A table that cannot be addressed by an exact single-column `id` primary key is not bulk-deleted when another table references it, or when the candidate rows are actually referenced through a self-referencing foreign key (declaring an unused self-FK in the schema is not enough). Cleanup leaves all matching rows in that table untouched and fails loudly with `table "<table>" is referenced by other tables but has no primary key of exactly the single column "id": cleanup can only SELECT/recurse by id, so this table's own descendants cannot be discovered and no rows were deleted for it` or `table "<table>" is referenced by itself via a self-referencing foreign key but has no primary key of exactly the single column "id": cleanup cannot safely bulk delete its rows because self-referencing descendants could be orphaned instead of removed`.
+
 ### Helpers
 
 - `requireId` validates every create-response id and every id passed to `track()` / `trackRow()`.
--- a/e2e-stack/docs/test-data.md
+++ b/e2e-stack/docs/test-data.md
@@ -170,7 +170,7 @@ the "level 50 but no limit granted" case instead.
 |                   |                                                                                       |
 | ----------------- | ------------------------------------------------------------------------------------- |
 | **Path**          | API `POST /v1/support/issue`                                                          |
-| **Returns**       | `{ supportIssueId?, uid, messageId? }`                                                |
+| **Returns**       | `{ supportIssueId, uid, messageId? }`                                                 |
 | **Options**       | `type` (default `GenericIssue`), `reason` (default `Other`), `name`, `message`, `tag` |
 | **Preconditions** | User must have **mail** (factory sets it if missing)                                  |
 
@@ -220,6 +220,10 @@ Deletes every row tracked during this process (`{ table, id }`) in reverse creat
 
 ### Helpers
 
+- `requireId` validates every create-response id and every id passed to `track()` / `trackRow()`.
+  It throws unless the value is a positive integer, so a successful API response with a missing
+  or invalid id cannot register a nonexistent cleanup row or let a test assert against data that
+  was never written.
 - `requireUnusedDeposit(blockchain)` — throws if no free deposit.
 - `ensurePersonalDataComplete(userDataId)` — SQL fill for `isDataComplete`.
 - `e2eMail(tag?)`, `TEST_IBAN`, `resetFactoryCounter()`.
--- a/e2e-stack/docs/test-data.md
+++ b/e2e-stack/docs/test-data.md
@@ -220,6 +220,8 @@ Deletes every row tracked during this process (`{ table, id }`) in reverse creat
 
 A table that cannot be addressed by an exact single-column `id` primary key is not bulk-deleted when another table references it, or when the candidate rows are actually referenced through a self-referencing foreign key (declaring an unused self-FK in the schema is not enough). Cleanup leaves all matching rows in that table untouched and fails loudly with `table "<table>" is referenced by other tables but has no primary key of exactly the single column "id": cleanup can only SELECT/recurse by id, so this table's own descendants cannot be discovered and no rows were deleted for it` or `table "<table>" is referenced by itself via a self-referencing foreign key but has no primary key of exactly the single column "id": cleanup cannot safely bulk delete its rows because self-referencing descendants could be orphaned instead of removed`.
 
+The self-reference check is deliberately conservative for a key spanning several columns: the catalogue carries no constraint identity to group pairs by, so each column pair is probed on its own. That can refuse a delete which the full key would have allowed, never allow one it would have refused.
+
 ### Helpers
 
 - `requireId` validates every create-response id and every id passed to `track()` / `trackRow()`.
--- a/e2e-stack/docs/test-data.md
+++ b/e2e-stack/docs/test-data.md
@@ -216,7 +216,7 @@ the "level 50 but no limit granted" case instead.
 
 ### `cleanupCreatedData()`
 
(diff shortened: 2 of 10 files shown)


TaprootFreak:
**EN:** Five review rounds were needed to reach zero findings, and the recurring lesson was that listing what filters a run is the wrong direction — the check is an allowlist now.
**DE:** Bis null Mängel brauchte es fünf Review-Durchläufe, und die wiederkehrende Lehre war, dass es die falsche Richtung ist, aufzulisten was einen Lauf filtert — die Prüfung ist jetzt eine Positivliste.

<details>
<summary>Details</summary>

## Review rounds

| Round | Lens | Findings | Outcome |
| --- | --- | --- | --- |
| 1 | Conformance | 9 | 7 fixed, 2 filed as #1315 and #1316 |
| 2 | Logic | 2 | both fixed |
| 3 | Both, end state | 5 | 4 fixed, 1 already covered |
| 4 | Both, end state | 3 | all fixed |
| 5 | Both, end state | 2 | 1 fixed, 1 confirmed as a deliberate choice |

Three rounds in a row found a flag missing from the list of things that count as a filter — `--project`, `--shard`, `--last-failed`, `--only-changed`, and then `--test-list` and `--ui`. Every omission let a partial run declare itself complete, which is the single failure this gate exists to prevent. Listing the filters can only ever be as current as the last time someone read the Playwright CLI. The check is an allowlist now: only flags known not to change *which* specs run leave a run complete, and anything unrecognised counts as a filter. That fails in the harmless direction. `--config` is off the list for the same reason — another config can set its own `testDir`, `testMatch` or project list, so a flag that can replace the whole selection cannot be one that promises the selection is complete.

## What the throw found

Making `cleanupCreatedData` throw instead of warn immediately showed that it had never cleaned up completely: six spec files failed on foreign key violations from rows the application itself writes and no test registers. Cleanup now descends the foreign keys, children first, and whether a table is a leaf is decided by whether another table references it — not by the shape of its primary key, and not by a self-reference.

## Not in this pull request

**#1315** — route coverage credits a splat claim for any navigation below it. **#1316** — the module-load wallet-counter seeding has no rejection handler. Both pre-existing.

## Verification

`npx tsc --noEmit` under the harness tsconfig on Node 20, and all seven CI checks green — including the full-stack E2E, which is the only thing that actually exercises the recursive cleanup and the claim-bound route gate against a real database.

TaprootFreak:
**EN:** Ready after four review passes, the last one clean in both dimensions.
**DE:** Bereit nach vier Review-Durchläufen, der letzte in beiden Dimensionen ohne Befund.

<details>
<summary>Details</summary>

Two passes per round, one on conformance and one on logic, repeated in full after every fix commit.
Thirteen findings, all fixed here except three filed separately.

**Round 1 (9 findings).** The foreign-key catalogue joined `key_column_usage` to
`constraint_column_usage` on the constraint name alone, which is a cross product for a key over
several columns — `(a, b) REFERENCES parent (x, y)` came back as four pairs, two of them
non-existent, and cleanup used each pair on its own to find rows to delete. Verified against a
throwaway Postgres: the old query returns `a→x, a→y, b→x, b→y`, the ordinality-joined one returns
`a→x` and `b→y`. Also in this round: a self-referencing table went through as a leaf, dead state in
`TableKeyInfo`, two comments stating the opposite of the code, and the claim-binding test proving
nothing.

**Round 2 (4 findings).** The fix for self-referencing tables was itself schema-based: a table that
merely *can* reference itself, holding one row whose self-reference is null, became unresolvable and
took the parent delete down with it. That is the same mistake this pull request set out to remove,
one branch over. It now probes whether a row actually references one of the rows about to be
deleted.

**Round 3 (2 findings).** Documentation gaps: why two symbols became exports, and where the
self-reference check is deliberately conservative for a key spanning several columns.

**Round 4.** No findings in either dimension.

**Verified rather than asserted.** The six new tests are confirmed executed and green in the
`Full-stack E2E` run for this head, not merely covered by a green job:

```
✓ trackRow rejects invalid ids before registering cleanup data
✓ getForeignKeys preserves composite foreign-key column ordinality
✓ cleanup recursively deletes untracked database children of a tracked user
✓ cleanup bulk-deletes a non-id-addressable child with an unused self foreign key
✓ cleanup rejects a non-id-addressable child with an actual self reference
✓ a claimed route opened only by another suite is reported as wrong-suite coverage
```

**Filed rather than fixed:** #1315, #1316, and the observation that a pure-function test sits in the
Playwright layer — `e2e-stack` has no layer below it, and introducing one would touch the very
project dependencies this pull request is repairing.

</details>

PR #1320, opened 2026-08-11
1962fdea - Let legacy Yapeal CHF holders switch between their old and new personal IBAN

TaprootFreak:
Follow-up to the CHF deposit cutover (#1305) and the API half DFXswiss/api#4813: customers who still hold an active legacy Yapeal CHF personal IBAN see their new Bank Frick IBAN by default and can flip back and forth with a provider toggle — with the guarantee that every PDF names exactly the IBAN that was on screen.

## Open, and not closable from this branch

- [ ] **DFXswiss/api#4813 merged and deployed first.** Not a hard ordering: without the new `bank` field on `GET /buy/personalIban` the switch offer never appears and this frontend behaves exactly as today (feature inert by construction). But the feature only exists once the API is live.
- [x] ~~**DFXswiss/packages#211 merged and released, then the dependency bump lands here.**~~ Done: the SDK is published as `@dfx.swiss/react` 1.8.0-beta.2 (pulling in `@dfx.swiss/core` 0.7.0-beta.1) and consumed from the registry — the local tarball override is gone and CI is green.

## What

1. **Frick by default for grandfathered holders.** The screens load the customer's personal-IBAN rows once (`GET /buy/personalIban`, now naming the issuing bank per row). An explicit decision tree — no fallback cascade — picks the provider: a toggle click always wins; an explicit URL/widget selector stays untouched; otherwise an existing still-payable Yapeal holder with KYC ≥ 50 gets `personalIbanProvider=Frick`; otherwise no selector is sent and the server default (the existing Yapeal row) applies. Below KYC 50 nothing changes at all.
2. **The provider toggle.** A second button next to the collection-IBAN toggle (distinct icon on purpose — two adjacent switches need distinct affordances) requests a fresh quote pinned to the other provider. Both directions are covered: a displayed Frick response offers the Yapeal row, a displayed Yapeal response offers Frick (KYC 50+ only).
3. **Display ↔ PDF consistency by construction.** Every flip is a new payment-info request with an explicit selector, and the API persists `bankId`/`virtualIbanId` on the transaction request (api#4813). The PDF button hangs off that request's id, so invoice and receipt rebuild from the stored IDs — the document can never name a different account than the screen did, in either toggle state.
4. **Visible, never silent, KYC fallback.** If the automatically chosen Frick default bounces with `KycRequired` (stale local KYC knowledge), the screen re-quotes without a selector and keeps a permanent hint ("we are showing your existing IBAN instead") — no blocking KYC screen for a choice the customer never made. An explicit toggle click or URL selector keeps the ordinary KYC error path.
5. **Selector plumbing generalized.** `normalizePersonalIban`/`toPersonalIbanProviderRequest`/`isUnrecognizedPersonalIbanSelector` now recognize every SDK enum member instead of hardcoding Frick, so `personal-iban=Yapeal` in the URL is an explicit (fail-closed, server-validated) selector instead of being silently dropped; the new `PersonalIbanProviderNotAvailable` token maps to customer-readable copy.
6. **The row list flows through the SDK, once, for both screens.** `GET /buy/personalIban` is consumed via `useBuy().getPersonalIbans` (DFXswiss/packages#211) inside a shared `usePersonalIbanRows` hook; the provider decision tree itself is a pure, unit-tested helper. The list is identity-tagged: an in-place account swap resets rows, override and fallback state (all of them — the KYC-fallback flag included), and neither a late list response nor a late *quote* response can commit for an account that is no longer signed in — the quote signature carries the customer identity, the committed payment info is stored together with the identity, method and the provider that was actually sent, and every commit re-checks the identity. KYC itself is evaluated only against a user object whose `accountId` matches the session identity; a foreign user counts as still loading. Completion and in-flight confirmation state reset on an identity change too, completion commits only on a *successful* confirmation (a rejected confirm surfaces as an error instead of a "Done!" screen), the confirm callback guards against unmount, and confirming is only possible once the final (exact-price) quote of a generation is displayed — each quote request persists its own transaction request, so this closes the window where a confirmation could target the approximate request while the screen moves on to the exact one. The displayed verification follows the sent provider (an explicit Yapeal answer after a URL-Frick entry is valid), the request provider derives from the same debounced snapshot as the request currency, the first quote additionally waits while the user context itself is loading, and both waits fail open after ten seconds so the core buy flow can never hang on the optional switch offer.
7. **The completion screen shows what was actually completed.** Both flows snapshot the completed payment information under the customer identity the moment the completion state is entered (after a successful confirmation on `/buy`, on the explicit "I have issued the transfer" action on `/buy/info`) and render only from that snapshot. Late-arriving context — a user reload, the personal-IBAN row list, a KYC re-derivation — can still re-quote in the background, but it can no longer swap the information a customer already acted on, neither on screen nor in the close payload handed to the host application. The snapshot is cleared together with the completion flag on an identity change. The `user` prop stays the raw context value on purpose: the completion component treats a missing user as "still loading" and then renders only a spinner while the screen suppresses the back button, so an identity-filtered `undefined` would strand a customer who has already paid — and that value gates nothing but the spinner and a copy variant, never account data.
8. **No dead ends.** If an explicitly selected provider is rejected by the server (`PersonalIbanProviderNotAvailable`), the error offers a "Show available IBAN" action that clears the selection and re-quotes selector-free — matching what the error copy promises, instead of a retry-only loop.
9. **Both buy flows.** `buy.screen` and `buy-info.screen` share the row hook and the derive helper and keep only their flow-specific wiring; `payment-info-buy` only renders the toggle.

## Tests

- Unit: the generalized selector utils, `getYapealAlternative` (bank/currency/active/acceptsPayments/status filters), `isVerifiedYapealPersonalIbanResponse`, and the new error-token mapping — full branch coverage on `personal-iban.ts`.
- The four screen suites mock the virtual-iban hook with a reference-stable, reset-proof interface: CRA sets `resetMocks: true`, which strips `jest.fn()` implementations before every test, and an unstable factory reference re-fires the load effect every render. Plain functions pinned inside the factory survive both; `createPersonalIban` rejects loudly if a suite ever reaches it.
- The KYC fallback renders as its own visual variant with committed baselines on both screens (first Frick request answered `KycRequired`, selector-less retry, permanent hint). The /buy/info suite carries the full scenario set (default Frick, toggle, fallback, hint clear, account swap, load rejection), and dedicated tests pin the same-snapshot, identity and combined user-timeout races with deferred promises and fake timers, plus completion-leak regressions on both screens.
- The close payload is pinned in the completion component's own suite: closing hands the host application exactly the payment information it was rendered with, so the chain screen -> snapshot -> host payload is covered end to end.
- e2e (static route mocks, deterministic): a legacy Yapeal holder sees the Frick IBAN first (request-body assertion on the sent selector), flips to the Yapeal IBAN (real production BIC `YAPECHZ2` in the fixture) and back; the provider-unavailable rejections run against fully static `/v2/user` fixtures (with the `accountId` decoded from the live session JWT, keeping the fixture coherent with the identity check) and the complete NestJS error shape, assert the new recovery action and click it, pinning the selector-free follow-up request in the request body. Eight committed baselines, each proven by a second run against the committed file, all visually verified.

## Per-file coverage and declared deviations

Measured over the full suite on this head (statements / branches / functions / lines):

|---|---|---|---|---|
| `src/util/personal-iban.ts` | 100 | 100 | 100 | 100 |
| `src/hooks/personal-iban.hook.ts` | 100 | 100 | 100 | 100 |
| `src/hooks/personal-iban-rows.hook.ts` | 100 | 96 | 100 | 100 |
| `src/hooks/virtual-iban.hook.ts` | 100 | 100 | 100 | 100 |
| `src/components/payment/payment-info-buy.tsx` | 57.14 | 92.59 | 40 | 55 |
| `src/components/payment/buy-completion.tsx` | 92.30 | 70 | 75 | 91.66 |
| `src/screens/buy.screen.tsx` | 85.60 | 81.22 | 69.01 | 87.33 |
| `src/screens/buy-info.screen.tsx` | 93.06 | 87.32 | 80.76 | 93.90 |

**Declared deviation (CONTRIBUTING § Coverage, reviewer's call):** the two screens and the payment-info component do not reach 100% on the whole file — the gap is dominated by pre-existing screen scaffolding this PR does not touch (the same state these files carried into #1305). The new decision tree, both toggle directions, the KYC fallback, the account-swap resets and the race guards are pinned by dedicated tests; a small remainder of defensive guard branches inside the new code is exercised only through the scenario suites rather than each by a dedicated test. Bringing an ~1100-line screen to 100% belongs to a preparatory coverage PR, not under this title.

**Declared deviation (§ full-stack test):** no `e2e-stack/` scenario is added here — the same declared deviation as in #1270, which is itself still open. The scenario needs a grandfathered Yapeal row *and* DFXswiss/api#4813 inside the stack image — until that API lands in the harness, the case is untestable there; the PR's own static Playwright suite pins both toggle states deterministically on `/buy` and `/buy/info`. It follows once the API half is merged into the stack.

## Pre-existing, reported not fixed

- `buy.screen.tsx` carries an empty `onSubmit` handler with a 2023 TODO ("fix broken form validation and onSubmit"); the actual submit path runs through the button handlers, so the dead handler has no runtime effect. It predates this PR (git blame 2023) and is reported here per the review rule rather than silently skipped; fixing the form wiring belongs to its own change.
- `useVirtualIban` still fires `POST buy/personalIban` through a raw `useApi().call` instead of the SDK, which CONTRIBUTING asks to avoid. It is pre-existing and identical on `develop`; this PR only narrows the file (the `GET` moved to `useBuy().getPersonalIbans`, and the response type now comes from the SDK). The remaining `POST` belongs to `personal-iban.screen`, not to the provider toggle, and migrating it means adding the endpoint to the SDK first — that would reopen the already-reviewed packages PR this branch is waiting on for its release. Reported here rather than silently skipped; it follows in its own change once the SDK release has landed.
- The sell and swap screens carry the same pre-existing two-phase quote commit (approximate, then exact price) with an unguarded confirm window; this PR closes that window for the buy screen it touches, the twin flows are untouched and keep their existing behavior.
- Three older cases in the two e2e files fail on a fresh local API stack independently of this branch (verified: identical failures on the `develop` merge-base against the same stack): the real-quote Frick issuance test needs a KYC ≥ 50 seed account and currently valid price rows, and the two Wallet-2 screenshots compare against baselines committed long before this branch that encode a different seeded account state. Fixing the seed or regenerating those baselines is environment work outside this diff.

## Evidence

Local chain on this head (Node 20, SDK tarball override, own API stack for the e2e auth): `lint`, `CI=true npm test` (84 suites / 1,052 tests), `build:dev`, `widget:dev`, index-swap check — all green. The eight provider e2e cases (`/buy` and `/buy/info`: both toggle directions, the KYC fallback and the provider-unavailable rejection) run deterministically against their eight committed, visually verified baselines; the two full spec files pass twice in a row with the identical result set (16 of 19, remainder = the three pre-existing environment cases above).

One test-only correctness find along the way: the static quote mocks omitted `asset.uniqueName`, which the buy screen's exact-price echo suppression compares — every static quote therefore triggered a redundant reverse quote cycle that could replace the payment info mid-interaction and reset the collection-IBAN toggle. All eleven static quote mocks now carry the field the real API always returns.


TaprootFreak:
**10 review passes** (conformance and logic, each pass covering the full diff) until both dimensions came back clean. Summary of what the later passes actually changed, since most of it is behaviour rather than polish:

**Cross-account and timing correctness**
- KYC is now evaluated only against a user object whose `accountId` matches the session identity; a foreign user counts as still loading and joins the existing ten-second fail-open. Previously an in-place session swap could let one account be judged with another account's KYC level.
- The automatic-Frick suppression flag was split from the KYC hint. Recovering from an unavailable provider no longer shows a misleading "requires KYC level 50" message; only a genuine `KycRequired` fallback renders it.
- Confirmation is pinned to the quote generation *and* to the displayed request id, and it is only possible once the final (exact-price) quote is committed. Each quote request persists its own transaction request, so this closes the window where a confirmation could target the approximate request while the screen moved on to the exact one. A rejected confirmation now surfaces as an error instead of a "Done!" screen.
- The completion screen renders from an identity-tagged snapshot of the completed payment information instead of the live quote state. Late-arriving context can still re-quote in the background, but it can no longer swap the information a customer already acted on — on screen or in the close payload handed to the host application.

**Dead ends removed**
- `PersonalIbanProviderNotAvailable` used to offer only a retry of the very provider that just failed. There is now a "Show available IBAN" action that clears the selection and re-quotes selector-free, matching what the error copy promises. Localized in all supported languages.

**Structure**
- The row-loading state machine moved into a shared hook and the provider decision tree into a pure, unit-tested helper, so both buy flows derive the provider from one place instead of two near-identical copies.

**Test findings worth naming**
- The static quote mocks omitted `asset.uniqueName`, which the exact-price echo suppression compares — every static quote fired a redundant reverse quote that could reset the collection-IBAN toggle mid-interaction.
- Two quote fixtures had IBAN values where one was a substring of the other; since the assertion checks containment, freezing the approximate instead of the final quote would have passed unnoticed. They are disjoint now.

**Deliberately declared, not fixed here** (see the PR description): the raw `POST buy/personalIban` in `useVirtualIban` (pre-existing, belongs to a different screen, needs an additive SDK change first), the pre-existing empty `onSubmit` handler, and the identical two-phase quote commit in the sell and swap screens.

One reported point was declined with reasons: an additional test combining the real completion component, a close click and an in-flight re-quote. Both ends of that chain are already pinned — the screen passing the frozen information (asserted during and after the background re-quote) and the component handing exactly its prop to the close channel (reference identity) — with nothing but prop forwarding in between.

This PR stays in draft until the SDK release it depends on has landed; the dependency bump follows, and CI is red by design until then.

<details>
<summary>Deutsch</summary>

**10 Review-Durchläufe** (Konformität und Logik, jeder über den vollständigen Diff), bis beide Dimensionen sauber waren. Was die späteren Durchläufe geändert haben, ist überwiegend Verhalten, nicht Kosmetik:

**Konto- und Timing-Korrektheit**
- KYC wird nur noch gegen ein Benutzerobjekt bewertet, dessen `accountId` zur Sitzungsidentität passt; ein fremdes Objekt gilt als „lädt noch" und läuft in das bestehende Zehn-Sekunden-Fail-open. Zuvor konnte ein Sitzungswechsel im laufenden Betrieb dazu führen, dass ein Konto mit dem KYC-Stand eines anderen bewertet wird.
- Die Unterdrückung des automatischen Frick-Defaults ist vom KYC-Hinweis getrennt. Nach einer Erholung aus einem nicht verfügbaren Provider erscheint kein irreführendes „erfordert KYC-Stufe 50" mehr; der Hinweis kommt nur noch beim echten `KycRequired`-Fallback.
- Die Bestätigung ist an die Quote-Generation *und* die angezeigte Request-Id gebunden und erst möglich, wenn die finale Quote (exakter Preis) steht. Jede Quote-Anfrage persistiert eine eigene Transaktionsanfrage — damit ist das Fenster geschlossen, in dem eine Bestätigung die approximative Anfrage treffen konnte, während die Anzeige längst auf der exakten stand. Eine fehlgeschlagene Bestätigung zeigt jetzt einen Fehler statt „Fertig".
- Der Abschlussbildschirm rendert aus einem identitätsgebundenen Abbild der abgeschlossenen Zahlung statt aus dem lebenden Quote-Zustand. Nachladender Kontext kann weiterhin im Hintergrund neu anfragen, aber nicht mehr die Information austauschen, auf die der Kunde bereits gehandelt hat — weder auf dem Bildschirm noch im Payload an die einbettende Anwendung.

**Sackgassen beseitigt**
- `PersonalIbanProviderNotAvailable` bot bisher nur einen erneuten Versuch mit genau dem Provider, der gerade fehlgeschlagen war. Jetzt gibt es „Verfügbare IBAN anzeigen": Auswahl zurücksetzen und selektorfrei neu anfragen — genau das, was der Fehlertext verspricht. In allen Sprachen übersetzt.

**Struktur**
- Die Zustandsmaschine für das Laden der Zeilen liegt jetzt in einem gemeinsamen Hook, der Entscheidungsbaum in einer reinen, unit-getesteten Funktion. Beide Kaufwege leiten den Provider aus einer Quelle ab statt aus zwei fast gleichen Kopien.

**Nennenswerte Testbefunde**
- Den statischen Quote-Mocks fehlte `asset.uniqueName`, das die Echo-Unterdrückung des exakten Preises vergleicht — dadurch löste jede statische Quote eine überflüssige Gegenanfrage aus, die den Sammel-IBAN-Umschalter mitten in der Bedienung zurücksetzen konnte.
- Zwei Quote-Fixtures trugen IBAN-Werte, bei denen der eine Teilstring des anderen war; da die Prüfung auf Enthaltensein testet, wäre das Einfrieren der approximativen statt der finalen Quote unbemerkt geblieben. Die Werte sind jetzt disjunkt.

**Bewusst deklariert, hier nicht behoben** (siehe PR-Beschreibung): der rohe `POST buy/personalIban` in `useVirtualIban` (Bestand, gehört zu einem anderen Screen, braucht zuerst eine additive SDK-Ergänzung), der vorbestehende leere `onSubmit`-Handler sowie der identische zweistufige Quote-Commit in den Verkaufs- und Tausch-Bildschirmen.

Ein gemeldeter Punkt wurde begründet abgelehnt: ein zusätzlicher Test, der die echte Abschlusskomponente, einen Klick auf Schliessen und eine laufende Neuanfrage kombiniert. Beide Enden dieser Kette sind bereits abgesichert — der Bildschirm übergibt das eingefrorene Abbild (geprüft während und nach der Hintergrund-Neuanfrage), und die Komponente reicht exakt ihr Prop an den Schliessen-Kanal weiter (Referenzidentität) — dazwischen liegt nichts als Prop-Durchreichung.

Der PR bleibt im Entwurf, bis das benötigte SDK-Release vorliegt; danach folgt die Abhängigkeitsanhebung. Bis dahin ist die CI absichtlich rot.

</details>

TaprootFreak:
**Ready for review.** The SDK dependency this branch was waiting on is published, so the last blocker is closed: `@dfx.swiss/react` 1.8.0-beta.2 (pulling in `@dfx.swiss/core` 0.7.0-beta.1) is now consumed from the registry in the pin style this repository uses, and the local tarball override is gone. Verified before bumping that the published packages actually carry the symbols this branch needs.

All checks are green on the final head, including the handbook image build that is skipped while a PR is a draft, and the PR reports mergeable. Local chain on Node 20 (the version this repository's CI uses): lint, 84 suites / 1,052 tests, `build:dev`, `widget:dev`, index-swap check — all clean; both Playwright spec files pass twice in a row with only the three pre-existing environment cases described in the PR body.

One remaining item in the checklist above is not closable from here: DFXswiss/api#4813 has to be deployed for the feature to exist at all. Until then this frontend behaves exactly as it does today — without the `bank` field on the row list there is no switch offer and no Frick default, so the change is inert by construction rather than by a flag.

<details>
<summary>Deutsch</summary>

**Bereit zum Review.** Die SDK-Abhängigkeit, auf die dieser Branch gewartet hat, ist veröffentlicht — damit ist der letzte Blocker erledigt: `@dfx.swiss/react` 1.8.0-beta.2 (zieht `@dfx.swiss/core` 0.7.0-beta.1 nach) wird jetzt in der hier üblichen Pin-Schreibweise aus der Registry bezogen, der lokale Tarball-Override ist weg. Vor dem Anheben wurde geprüft, dass die veröffentlichten Pakete die benötigten Symbole tatsächlich enthalten.

Alle Checks auf dem finalen Stand sind grün, inklusive des Handbook-Image-Builds, der bei Entwürfen übersprungen wird; der PR ist mergefähig. Lokale Kette auf Node 20 (der Version, die die CI dieses Repositories nutzt): lint, 84 Suiten / 1'052 Tests, `build:dev`, `widget:dev`, Index-Swap-Prüfung — alles sauber; beide Playwright-Dateien laufen zweimal hintereinander durch, mit ausschliesslich den drei vorbestehenden Umgebungsfällen aus der PR-Beschreibung.

Ein Punkt der Liste oben lässt sich von hier aus nicht schliessen: DFXswiss/api#4813 muss deployt sein, damit das Feature überhaupt existiert. Bis dahin verhält sich dieses Frontend exakt wie heute — ohne das `bank`-Feld in der Zeilenliste gibt es kein Umschaltangebot und keinen Frick-Default, die Änderung ist also konstruktionsbedingt wirkungslos statt über einen Schalter.

</details>

PR #1130, opened 2026-06-19
feat(support): support dashboard overview with triage and statistics

joshuakrueger-dfx:
## Support dashboard — overview

A triage-focused landing for the support dashboard. Opening "Support Dashboard" from the burger menu now shows this overview; the full ticket list moves to `/support/dashboard/all`.

### What's included
- **My tickets** — tickets assigned to the logged-in agent (resolved from the backend `support/issue/clerk` mapping, no localStorage), shown as *assigned / open* on one card.
- **Waiting longer than** — customer-waiting tiers `1h / 12h / 24h` (24h = escalated). The pills are clickable filters that drive a time-sorted list (longest wait first); 24h shows the escalations.
- **Statistics tab** — selectable period (7d / 30d / 6m / 12m): new tickets, avg messages per ticket, tickets/day, and a day/month trend chart. Backed by the new `support/issue/statistics` aggregate with a client-side fallback.
- Pure aging/escalation/statistics logic extracted to `src/util/support-stats.ts` (no React deps) with unit tests.

### Notes
- Colour is semantic only: red = escalation (≥24h waiting), amber = 12h tier; everything else neutral.
- German translations added (consistent with the existing partial fr/it coverage).
- Identity comes from the backend account→clerk mapping (see DFXswiss/api#3942). If an account isn't mapped, "My tickets" shows a hint instead.

### Checklist
- [x] tsc clean, ESLint clean
- [x] Unit tests (support-stats) green
- [ ] Smoke-test on dev with a support/admin login

> Pairs with DFXswiss/api#3942 (backend). Works before it deploys via the client-side stats fallback; merge the backend first for accurate numbers and clerk resolution.

PR #1321, opened 2026-08-11
0257fbbe - Add a staff view for unassigned bank transactions

TaprootFreak:
## Why

Bank transfers that the automation could not resolve sit in the database as `Pending`, `GSheet` or `Unknown`, and someone has to decide what they are. Customers can already resolve their own through `/transaction/unassigned`, but the rest — the ones nobody claims — had no view in this app at all. Staff could only reach a bank transaction by searching for an IBAN or a name they already knew, which is why the leftovers were worked through a Google Sheet.

This adds the missing screen. It consumes `GET /v1/bankTx/unassigned` from DFXswiss/api#4825 (merged).

## What

`/compliance/bank-tx/unassigned` lists the open bank transactions of the last three months with amount, direction, name, IBAN, remittance info and the buy route the payment reference points at.

- **Support, Compliance and Admin all see the list**, through a new `useSupportStaffGuard`.
- **Only Admin can assign.** The type dropdown, the buy-ID field and the save button render under `session?.role === UserRole.ADMIN`. This is presentation, not authorisation — `PUT /v1/bankTx/:id` carries `RoleGuard(UserRole.ADMIN)` on the server, so the read-only view holds even for someone who reaches the route directly.
- **Debits are marked.** A debit can never become a `BuyCrypto` — the API rejects it — so the direction column highlights `DBIT` rather than letting it read like an incoming payment. The API correspondingly proposes a buy candidate only for credits.
- **The buy candidate pre-fills the field** but is never submitted on its own; a human picks the type and confirms.
- **A save locks the row until the list has refreshed.** All rows share one form instance, so releasing the lock at the PUT response would leave the just-assigned row clickable while the reload was still in flight — long enough to assign it twice.
- The buy ID is validated lexically before `Number()` touches it: the conversion is lossy, and an oversized value would silently land on a different route's id.
- The type list deliberately omits `Pending`, `GSheet` and `Unknown`: putting a transaction back into the unassigned state is not an assignment, and the API maps those three to a null transaction type.

## Coverage

Per file, measured with `--collectCoverageFrom` on each:

| `src/hooks/bank-tx-assignment.hook.ts` | 100 | 100 | 100 | 100 |
| `src/screens/compliance-bank-tx-unassigned.screen.tsx` | 100 | 100 | 100 | 100 |
| `src/hooks/guard.hook.ts` | 100 | 100 | 100 | 100 |
| `src/components/compliance/quick-links-section.tsx` | 100 | 100 | 100 | 100 |

Suite: 85 files, 1027 tests, green. The new API calls live in their own small hook rather than being appended to `src/hooks/compliance.hook.ts` (1581 lines), which stays untouched.

## Handbook and visual regression

Three baselines for the new screen, covering each visual variant it introduces: the list with the highlighted debit row, the expanded assign form (buy candidate pre-filled, save disabled until a type is picked), and the read-only view a non-admin sees. Plus a `scripts/handbook/metadata.json` entry.

Two existing baselines were regenerated because this PR changes what they show:

- `support-dashboard-01-overview` — the new navigation entry. Regenerating it surfaced a pre-existing defect: the spec pointed at `/support/dashboard`, but the screen it describes moved to `/support/dashboard/all` in #1130, so it had stopped covering that screen entirely. Route corrected in this PR.
- `compliance-02-search-results` — the new quick link. That baseline dated from #919 and predated the quick-links section itself (#1081), so it was stale by a whole section.

## Full-stack E2E

The route is claimed in `e2e-stack/specs/registry/compliance.ts` and visited by a test in `compliance-cases.spec.ts` that creates a pending credit, opens the screen as Admin, then re-opens it as Compliance to assert the assign action is absent.

## Declared deviations

**1. `src/App.tsx` (669 lines) and `src/screens/support-dashboard.screen.tsx` (480 lines) do not reach 100 %.** Each receives a handful of lines — a lazy import plus a route entry, and one navigation button. Bringing either whole file up is an order of magnitude more work than the change itself.

**2. The two endpoints are called through `useGuardedApi` rather than added to the SDK first.** `CONTRIBUTING.md` says a missing endpoint should be added additively to `DFXswiss/packages`, released, and consumed from here. `@dfx.swiss/react` covers neither of these endpoints today, and nine existing hooks under `src/hooks/` — `compliance`, `dashboard`, `scorechain`, `support-templates`, `realunit-api`, `realunit-support`, `realunit-compliance` among them — call internal staff endpoints exactly this way. Following the rule literally means an SDK change in another repository plus a release that this PR would then depend on. That is a call for the reviewer, not one to make silently, so it is declared here rather than worked around.


TaprootFreak:
**EN:** Ready after five review passes across two dimensions, conformance and logic, iterated to zero findings.
**DE:** Bereit nach fünf Review-Durchläufen über zwei Dimensionen, Konformität und Logik, iteriert bis null Befunde.

<details>
<summary>Details</summary>

Two independent reviews per pass — one against `CONTRIBUTING.md`, one on logic and correctness — repeated in full after every fix, with the expected HEAD pinned so a stale pass would abort rather than pass quietly.

**Pass 1** — three fixes:

- The support dashboard admits Marketing, but the target route does not; the new link was ungated, so Marketing would have been bounced straight back.
- Handbook coverage was missing entirely for the new screen — no baseline, no metadata entry.
- The logic pass was inconclusive and repeated.

**Pass 2** — three fixes:

- Switching rows during an in-flight save would have applied that save's outcome to the wrong row, since all rows share one form instance. Row buttons are now locked while saving.
- The buy ID was converted with `Number()` before being checked for integrality. The conversion is lossy: `1.0000000000000001` becomes `1`, and an oversized value rounds onto a different route's id. Validation now happens on the text first.
- A test comment described an early return the handler does not have.

**Pass 3** — three fixes:

- Baselines for the compliance search page were stale, dating from #919 and predating the quick-links section (#1081) they were supposed to show.
- The row lock and the safe-integer bound were both unverified — either could have been removed without a test turning red.
- A mock for a request no test ever issued was removed; an unexpected request now surfaces instead.

**Pass 4** — two fixes:

- After a successful save the list reload was not awaited, so the lock lifted while the refresh was still running and the just-assigned row stayed clickable.
- The screen has two visible role variants, but only the admin one had a baseline.

**Pass 5** — no findings in either dimension.

One reported point is not adopted but declared in the description instead: routing the two calls through the SDK would require an additive change and release in another repository. Nine existing hooks here call internal staff endpoints the same direct way, so the trade-off is the reviewer's to make.

Two pre-existing defects were fixed along the way rather than left in place: the support dashboard spec had been pointing at the wrong route since #1130 and therefore covered nothing, and the compliance search baseline had been stale by an entire section.

</details>

PR #1425, opened 2026-08-24
fix(compliance): let a Repeat call outcome release the transaction

Blume1977:
EN:
Repeat now releases this one transaction through the new API endpoint, writes no check date on the account, and no longer offers an AML action. Pass stays Admin-only; the pipeline still decides the verdict. Merge after the companion API change is on the development environment, or the new endpoint 404s.

DE:
Repeat gibt jetzt diese eine Zahlung über den neuen API-Endpoint frei, schreibt kein Check-Datum auf dem Konto und bietet keine AML-Aktion mehr an. Pass bleibt Admin-only; die Pipeline entscheidet weiterhin das Verdikt. Erst mergen, wenn die zugehörige API-Änderung auf der Entwicklungsumgebung liegt, sonst 404t der neue Endpoint.

<details>
<summary>Details</summary>

## Problem

Repeat means: let this payment through, call the customer again next time. The tool wrote the phone-call status and left the transaction to the clerk — and none of the offered AML actions could finish the job:

- Pass is Admin-only (the API rejects it for Compliance), and the option is already hidden for non-Admins.
- Reset re-runs the very check the deliberately missing check date is failing, so the payment loops straight back into the queue it came from.
- Fail is wrong.

## Solution

Repeat now decides the transaction itself, exactly like Completed and Failed do:

- The save releases this one transaction through `PUT :id/phoneCallCleared`, which clears the call-queue holds for that transaction only.
- The AmlCheck selector disappears for Repeat. Offering it would only invite the combination that cannot work.
- The release is written before the implied reset — the API accepts it only while the transaction still carries its queue reason, which the reset clears.

The account is untouched apart from its status: no check date is written, so its next transaction is held again and the customer reappears in the call list. The verdict stays with the API — a released transaction passes only when no other AML error remains.

The implied reset is unchanged in its own right: still only for the queues the API excludes from its recheck cron (`ManualCheckIpCountryPhone`), still Reset rather than Pass/Fail.

## Failure behaviour

A rejected release fails the save with `failedStep: 'transaction'` and no KYC log is written. Reporting success there would tell the clerk the payment is on its way while it silently stays in the queue.

## Tests

- Write order pinned: `userData` → `phoneCallCleared` → `reviewReset` → log, and no check date in the userData payload.
- Sell transactions go through `buyFiat/:id/phoneCallCleared`.
- A queue entry without a transaction still writes nothing but the log on Repeat.
- A rejected release surfaces as a failed transaction step with no log entry.
- Form: no AmlCheck selector on Repeat; automatic Reset on the recheck-blocked queue and none where the cron re-evaluates anyway.

</details>

### DIFF
--- a/src/__tests__/call-queue-outcome-form.test.tsx
+++ b/src/__tests__/call-queue-outcome-form.test.tsx
@@ -277,6 +277,35 @@ describe('CallQueueOutcomeForm AmlCheck action', () => {
   it('defaults an open-ended outcome to no change', async () => {
     renderForm(TX_CONTEXT);
 
+    fillAndSubmit(CallOutcome.SUSPICIOUS);
+
+    await waitFor(() => expect(mockSaveCallOutcome).toHaveBeenCalledTimes(1));
+    expect(submittedAmlAction()).toBeUndefined();
+  });
+
+  // Repeat decides the transaction on its own: the hook releases this one transaction and the account
+  // stays in the queue for its next one. Asking for an AML action on top would only offer the plain
+  // reset, which re-runs the very check the still-missing check date is failing.
+  it('asks for no AML action on Repeat', () => {
+    renderForm(TX_CONTEXT);
+
+    fireEvent.change(screen.getAllByRole('combobox')[0], { target: { value: CallOutcome.REPEAT } });
+
+    expect(screen.getAllByRole('combobox')).toHaveLength(1);
+  });
+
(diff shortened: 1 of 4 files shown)


TaprootFreak (approved):
EN:
Looks correct: Repeat now releases this transaction through the new endpoint, without writing a check date and without offering Pass.

DE:
Stimmt: Repeat gibt diese Zahlung über den neuen Endpoint frei, schreibt kein Check-Datum und bietet kein Pass an.

PR #1442, opened 2026-09-01
01a0392a - test(e2e-stack): cover the self-custodial Lightning login and the sell route behind it

joshuakrueger-dfx:
## Open, and not closable from this branch

- [ ] **This pull request cannot demonstrate its own specs, and no label changes that.** `e2e-stack.yml:21-24` states it and `:105-108` implements it: a fork pull request has no `E2E_API_CHECKOUT_KEY`, so the job takes `mode=none` and the stack never comes up — with `ci`, with `ci:full`, before or after DFXswiss/backend#5331 lands. To run these specs the branch has to live in this repository rather than in a fork, which needs write access the author does not have. **Merge DFXswiss/backend#5331 first**; from then on that repository's own Full-stack E2E job exercises them on every run, because it takes the harness from `app@develop`. The order is not a preference: this harness builds the API from `backend@develop` (`pr.yml` never passes `api_ref`, so `e2e-stack.yml` falls back to `develop`; a real job log shows `Syncing repository: DFXswiss/backend` … `+refs/heads/develop*` … `checkout --force -B develop`). Before that change has landed, the two Lightning specs here fail with the 400 they exist to demonstrate. Applying the label earlier would produce a red run, not a measurement. Without it a fork head gets no run: all five checks report *skipped*, so nothing here has been measured by CI. Applying it from this account fails with `does not have the correct permissions to execute AddLabelsToLabelable`, and `.github/workflows/ci-on-ready.yaml` states the same for the Ready path in its own words: `fork draft — GITHUB_TOKEN cannot workflow_dispatch a fork head` / `a human must apply the ci label so the labeled event starts CI`. Setting this pull request ready would therefore not start CI — it would fail that job and convert the pull request back to draft. **An administrator of DFXswiss/app has to apply the label.** Everything the label unblocks was measured locally instead; the numbers and their limits are in the closing comment.
- [ ] **Re-measure the full-stack suite figures in `docs/test-architecture.md`** once that run exists. The section defines them as a CI measurement, so a local number taken under a substituted frontend image would be a different figure wearing the same name. This is the declared deviation named below.

**Symptom (verbatim):** `POST /v1/auth failed: 400 {"statusCode":400,"message":"Invalid signature","error":"Bad Request"}` — measured in this harness against an API image built from the API repository's `develop`, using a Lightning address and a 128-character hexadecimal signature.
**Scale:** Every sign-in of a self-custodial Lightning wallet. No spec in this harness has ever posted a Lightning address — `specs/fixtures/auth.ts` signs in with `ethers` EVM keys only, so a green full-stack run says the old login still works and nothing about the new one.
**Declared deviation:** `docs/test-architecture.md` defines the full-stack suite measurement as a CI measurement and asks for a re-measure after any change to the suite. This pull request does not carry one: a fork head gets no run here without the `ci` label, which only an administrator can apply. The figures in that section are therefore left untouched rather than filled with a local number measured under a substituted frontend image.
**Smaller fix considered:** Asserting only that the API accepts the signature shape — rejected, because it cannot tell the two API versions apart: the request DTO already accepts 128 hexadecimal characters through unrelated chain alternatives, so such a test passes against the old API too.
**Skipped check (K1.6a):** No check has run on this pull request, and none can be started from this branch. CI here is kicked by `.github/workflows/ci-on-ready.yaml` on `ready_for_review`, and a fork head additionally needs the `ci` label; applying it fails with `joshuakrueger-dfx does not have the correct permissions to execute AddLabelsToLabelable`. The five visible checks are therefore *skipped*, not failed — and neither classic branch protection nor a ruleset on `develop` marks any of them required, both queried and both empty.

**Skipped check (K1.13):** Same cause as K1.6a. The six workflows whose path filters the changed files would trigger never started, for the one reason above — an administrator has to apply the `ci` label. This is carried as an open task at the top of this description rather than argued away.

**EN:** Tests only, no product code. Two specs sign a self-custodial Lightning wallet in against the real API and a real Postgres, then follow that session to a sell payment info — deposit address, quote and a payment request. Against an API image without the corresponding API change both fail with the 400 above; with it they pass.

## What a green run here does NOT prove

Two limits, stated because a reader would otherwise assume more:

- **No payment is proven.** The harness sets `DISABLED_PROCESSES=*`, so pay-in detection, AML, purchase, liquidity, payout and ledger booking never execute. This layer verifies the synchronous path up to the payment request and nothing beyond it. That is the harness's documented boundary, not a gap this branch introduces.
- **The invoice comes from a stand-in.** `e2e-stack/images/lnurl/` answers the two calls the API makes and checks that the callback is asked for the expected amount, but it returns a fixed invoice — a vector from the BOLT-11 specification. The run therefore says nothing about node interoperability, routability or payability, and the assertion checks the mainnet prefix and the bech32 checksum, not the invoice's mandatory fields.

<details>
<summary>Details</summary>

## Why there is an LNURL stand-in, and why it answers on `localhost`

The API only produces a payment request when the quote is valid, and for Lightning it resolves the deposit LNURL and fetches an invoice — two real outbound HTTP calls. The deposit address is gated the same way. Without a counterpart the sell call ends in an error, so the path cannot be exercised at all.

The obvious placement — a service on the `sandbox` network under its own hostname — does not work, and the reason is worth recording for whoever reads this next. In `loc`, which this stack runs, the API intercepts **every** outbound HTTP call and answers with a canned response; the only exemption is a URL containing `localhost` or `127.0.0.1`. A stand-in under any other hostname is therefore never reached. It shares the API container's network namespace (`network_mode: "service:api"`) and answers on `localhost`, so the real code path runs instead of a canned reply — without touching API production code.

It carries `restart: unless-stopped` because `up.sh` restarts the API during seeding, and a sidecar sharing that namespace does not come back on its own.

## Measured

Against an API image built from the branch that adds the self-custodial Lightning login:

```
✓ auth.spec.ts › new wallet receives a token and stores its exact signature   (2.6s)
✓ auth.spec.ts › same wallet and signature receive another token              (26ms)
✓ auth.spec.ts › same wallet with a different signature receives 401          (27ms)
✓ sell-lightning.spec.ts › returns the configured LNURL and a mainnet-prefixed, checksum-valid payment request (4.8s)
```

Against an API image built from that repository's `develop` — change absent, verified by counting zero occurrences of the new signature format in both production files — the same specs fail:

```
✘ auth.spec.ts › new wallet receives a token and stores its exact signature
✘ sell-lightning.spec.ts › returns the configured LNURL and a mainnet-prefixed, checksum-valid payment request
  Error: POST /v1/auth failed: 400 {"statusCode":400,"message":"Invalid signature"}
```

## Not verified here

The fourth auth case — `Lightning access token authenticates the frontend navigation` — is the only one that needs the real frontend image. It was **not** verified locally: that image could not be built on the machine used here, because the frontend build exceeds the Docker VM's memory. The local run substituted a placeholder, so the case is red for that reason alone. It is expected to run in CI, where the prebuilt frontend images are pulled.

## Review findings closed

Three review passes read this branch against the repository's own rules; the last one reported no
findings. Closed along the way: the stand-in was bind-mounted where the harness documents it has
none, so it builds an image and `up.sh` recreates it on a warm stack; users created through the real
login were never registered for cleanup in a shared, serially-used database; the reality declaration
listed neither the two SQL shortcuts nor the synthetic signature the fixture generates; the callback
accepted any amount; an assertion pinned a constant of the encoder rather than anything the API
said; the fiat precondition filtered `buyable` where a sell needs `sellable`; two documentation
figures had gone stale; and a comment on the compose service was German in an English public
repository.

## Earlier findings

Two independent review passes read this branch against the repository's own rules. Eight findings,
all closed: the stand-in was bind-mounted where the harness documents that it has none, so it now
builds an image; users created through the real login were never registered for cleanup in a shared,
serially-used database; the reality declaration listed only the stand-in and not the two SQL
shortcuts; the callback accepted any amount; two assertions could not fail; the sell test name
claimed more than it checked; and two figures in the documentation had gone stale.

## The sign-in fixture

The fixture posts `{ address, signature }` to the same endpoint `signatureLogin` uses. It does **not** sign the server's message: the signature is a synthetic 128-character lowercase hexadecimal value, generated once per wallet and then reused deliberately, because the re-login case has to send the identical value. A green run therefore proves nothing about real Breez/Spark signing or about that SDK's determinism — only that the API accepts such a value as a credential and holds it.

A Lightning address carries no key of its own, so this signature cannot be verified cryptographically. The API treats it as a shared secret, the same way it already treats the provider-issued ownership proof of a custodial Lightning wallet: the first sign-in establishes it, every later one has to repeat it. The third case is what makes that meaningful — a different signature for the same address is refused.

## Deposit fixture

`createLightningDeposit` follows the existing pattern used by `createPaymentLink`: prefer a free Lightning deposit, otherwise insert a synthetic one by SQL. The address is a bech32 LNURL derived from the stand-in's URL at seed time rather than hardcoded, and the spec asserts once that it decodes back to that URL.

</details>


joshuakrueger-dfx:
EN:
Ready after 3 review passes.
Cover the self-custodial Lightning login and the sell route behind it against the real API and a real Postgres.

<details>
<summary>Details</summary>

Head `42c4f875f4`, eighteen files, +458/−8. Tests only — no product code.

### What each pass closed

**Pass 1** — the LNURL stand-in was bind-mounted, which `e2e-stack/README.md` states this harness does not do and calls unsupported for the environments it needs; it now builds its own image. Users created through the real login were never registered for cleanup, in a database that is shared and used serially. The reality declaration listed only the stand-in, not the two SQL shortcuts. The callback accepted any amount. Two assertions could not fail. A test name claimed a BOLT-11 payment request where the assertion checks the prefix and the checksum. Two documentation figures had gone stale.

**Pass 2** — an assertion pinned a constant of the encoder rather than anything the API returned. The fiat precondition filtered `buyable` where a sell needs `sellable`; CHF is both in the seed, so the case stayed green while checking the wrong column. A comment claimed a mainnet check the function does not perform. The compose comment was German in an English public repository. The new factory was missing from the catalogue that lists every other one.

**Pass 3** — the synthetic signature was not declared: the fixture generates a random value and never signs the message it fetched, which is the right shape for this layer but had to be said. And `up.sh` built the lnurl image while the following `compose up` did not name the service, so a reused stack kept the old one — the promise in the comment was false. Both closed; the pass after that reported no findings from either dimension.

### Measured

Against an API image built from the branch that adds the self-custodial Lightning login:

```
✓ auth.spec.ts › new wallet receives a token and stores its exact signature
✓ auth.spec.ts › same wallet and signature receive another token
✓ auth.spec.ts › same wallet with a different signature receives 401
✓ sell-lightning.spec.ts › returns the configured LNURL and a mainnet-prefixed, checksum-valid payment request
```

Against an image built from that repository's `develop` — the change absent, verified by counting zero occurrences of the new signature format in both production files — both Lightning specs fail at the first call:

```
Error: POST /v1/auth failed: 400 {"statusCode":400,"message":"Invalid signature"}
```

### Not verified, and why

- **The browser navigation case is red locally.** It is the only one needing the real frontend image, which could not be built on the machine used here — the build exceeds that Docker VM's memory. The run substituted a placeholder, so the case fails for that reason alone. It is expected to pass in CI, where the prebuilt frontend images are pulled.
- **No CI run on this pull request.** A fork head gets no checks here without the `ci` label, which only an administrator can apply. The five checks therefore show as skipped. Neither classic branch protection nor a ruleset on `develop` marks any of them required — both were queried and both are empty — so this is documented rather than worked around.
- **The suite measurement in `docs/test-architecture.md` is not re-measured**, for the same reason: that section defines it as a CI measurement, and a local number taken under a substituted frontend image would be a different figure wearing the same name. This is the declared deviation named in the description.

### Open comments

None. No review threads, no inline comments, no change requests.

### Mergeability

`mergeable: MERGEABLE`.

</details>

PR #1483, opened 2026-09-11
28835ea3 - Only auto-connect wallets that are available

Daniel-DFX:
EN:
Wallet connect screens no longer start a connect attempt for a wallet that is not available in the browser; that attempt failed behind the install hint and was reported as a client error. Because the availability check now decides whether a connect happens, each hook detects its wallet the way its own connect path does: through the adapter's state for Phantom and Trust on Solana, through a short wait for a wallet injected after the page loaded for Alby and the two Tron wallets, and through the mobile redirects into the wallet apps wherever the adapter performs them. When the page is sent to a wallet app and the app does not take over, the screen returns to the wallet selection instead of showing and reporting an error, and an unmounted ConnectBase starts no login and reports nothing for that attempt, so a wallet approved after the user moved on can no longer log that wallet in. This PR declares a deviation from the handbook rule, explained in the details.

DE:
Die Wallet-Connect-Screens starten keinen Verbindungsversuch mehr für eine Wallet, die im Browser nicht verfügbar ist; dieser Versuch schlug hinter dem Installationshinweis fehl und wurde als Client-Error gemeldet. Weil die Verfügbarkeitsprüfung jetzt über den Connect entscheidet, erkennt jeder Hook seine Wallet so wie sein eigener Connect-Pfad: über den Adapter-Zustand bei Phantom und Trust auf Solana, über ein kurzes Warten auf eine nach dem Laden injizierte Wallet bei Alby und den beiden Tron-Wallets und über die mobilen Weiterleitungen in die Wallet-Apps, wo der Adapter sie ausführt. Wird die Seite an eine Wallet-App übergeben und die App übernimmt nicht, kehrt der Screen zur Wallet-Auswahl zurück, statt einen Fehler anzuzeigen und zu melden, und ein unmountetes ConnectBase startet keinen Login mehr und meldet für diesen Versuch nichts, sodass eine nach dem Weggehen freigegebene Wallet nicht mehr eingeloggt werden kann. Dieser PR deklariert eine Abweichung von der Handbook-Regel, begründet in den Details.

<details>
<summary>Details</summary>

## Problem

`ConnectBase.init()` computed `supported = await isSupported()`, showed the install hint when it was false, and then called `connect()` anyway whenever `autoConnect` was set:

```ts
if (autoConnect) connect();
```

For MetaMask without an extension, `new Web3(Web3.givenProvider)` has no provider, `web3.eth.requestAccounts()` throws web3's `Provider not set or invalid`, `connect()` stores it as the connect error, and `ConnectError` reports it through `useReportDisplayedError` as a `HandledError`, although the user only sees the install hint (the content with the error is rendered hidden behind it). The same applies to every other wallet component whose availability check can come back false while `autoConnect` is set, each with its own SDK's error text. Taro and WalletConnect set `autoConnect` unconditionally as well, but their `isSupported` is always true, so the pattern never arose there.

### `connect-base.tsx`

```ts
if (autoConnect && supported) connect();
```

The `autoConnect` props of the wallet components stay unconditional on purpose: gating them on `activeWallet === props.wallet` (like BitBox, CLI, Ledger, Trezor) would also stop auto-connecting on a wallet switch, which is a different behavior.

### Wallet detection reviewed

Before this change, a wallet that `isSupported()` missed could still connect, because the hidden `connect()` ran anyway — including the waits the wallet SDKs perform internally. With the gate that path is gone, so each `isSupported()` has to answer "can this wallet's connect path succeed from here", not just "is some global set at this instant".

| Component | `isSupported` before | Result |
| --- | --- | --- |
| Taro | `() => true` | Correct: LNURL login via QR code or app link, no browser wallet involved. Unchanged. |
| WalletConnect | `() => true` | Correct: QR code or deep link, no browser wallet involved. Unchanged. |
| Trust (Solana) | `window.ethereum?.isTrustWallet` | **Fixed.** Checks the EVM provider, while the adapter connects through `window.trustwallet.solana`; false whenever another extension (MetaMask, Rabby) owns `window.ethereum`. Now: adapter `readyState` is `Installed`. The adapter's `connect()` does not wait for a late wallet, so a check at mount matches it. |
| Phantom | `window.phantom?.solana.isPhantom` | **Fixed.** Throws a `TypeError` when `window.phantom` has no `solana` provider (the promise in `init()` then rejects and the spinner never ends), ignores `window.solana`, and is false on iOS Safari, where the adapter opens the page in the Phantom app. Now: adapter `readyState` is `Installed` or `Loadable`. |
| Alby | `Boolean(window.webln)` | **Fixed.** `enable()` waits up to ~100 ms (`waitForWebln`) for a late injected `window.webln`, but that wait was only reachable through `connect()`. Now `isSupported` is the new `useAlby().isAvailable()`, which uses the same wait. |
| Trust (Tron) | `window.ethereum?.isTrustWallet` | **Fixed.** Same EVM check, while the adapter connects through `window.trustwallet.tronLink`, waits for a late wallet, and on a mobile browser outside the Trust app opens the page in the app. Now: `isInstalled()` is `supportTrust() \|\| isTrustApp() \|\| (isInMobileBrowser() && !isTrustApp())`, and `isSupported` is the new `isAvailable()`, which retries it. |
| TronLink | `Boolean(window.tronLink)` | **Fixed.** The adapter also accepts `window.tron` (TIP-1193) and `window.tronWeb`, waits for a late wallet, and on a mobile browser outside the app opens the page in the app. Now: `isInstalled()` is `supportTronLink() \|\| isInTronLinkApp() \|\| (!supportTronLink() && isInMobileBrowser() && !isInTronLinkApp())`, and `isSupported` is the new `isAvailable()`. |

`WalletReadyState` and `isInMobileBrowser` come from the adapters' base packages, which are installed as their dependencies; `src/hooks/tron.hook.ts` already imports from `@tronweb3/tronwallet-abstract-adapter` the same way.

### Waiting for a wallet that is injected late

Both Tron adapters tolerate a wallet that appears after the page loaded: their constructor starts `_checkWallet()` and `connect()` awaits it, polling every 100 ms. Since the gate decides once at mount, the hooks now do the same through `isAvailable()`, which retries `isInstalled()` every 100 ms — the delay-retry shape `alby.hook.ts` already uses for `window.webln` — and checks once more after the last wait, so the closing window is not missed.

The wait is deliberately capped at 2 s for both, instead of following the adapters' own `checkTimeout` (2 s for Trust, **30 s** for TronLink): the check runs before anything is shown, so the cap is what a user without the extension waits before seeing the install hint. Two seconds matches the Trust adapter's own default and covers a slow content script; a 30 s spinner would not be acceptable for the far more common case of a wallet that is simply not installed.

The cap is therefore a deliberate trade-off, not full parity with the adapter: a TronLink wallet that only appears between 2 s and 30 s after the screen opened is shown as unavailable, although the adapter alone would still have found it. Reopening the wallet re-runs the check, and the install hint is the honest answer for everything slower than the cap.

The Trust adapter needed one more thing: it runs its own readiness poll from its constructor and caches the verdict, with a default window of 2 s — the same length as ours, but started earlier, during render rather than in the effect. Its cached "not found" could therefore be older than our own decision, and `connect()` would answer from that cache: open trustwallet.com in a new tab and throw. It is now constructed with `checkTimeout: 3000`, so its window always outlives ours. TronLink needs no such change; its default is 30 s.

### Redirects into a wallet app

On three paths the adapter does not connect but navigates the page into the wallet app: Phantom on iOS Safari (`connect()` sets `location.href` and returns without a public key), and Trust (Tron) / TronLink on a mobile browser outside their app (`connect()` sets `location.href` and throws `WalletNotFoundError`). The hooks surfaced these as `No public key found` / `The wallet is not found.`, which `ConnectError` showed and reported. The hooks now wait 5 s and throw `AbortError`, the pattern `ConnectAlby` already uses after its own redirect; `ConnectBase` maps it to `onCancel`, which only resets the wallet selection state. For the two Tron wallets the redirect decision is taken before the adapter is called, the same moment the adapter itself decides, so a failure that happens for any other reason is never reclassified as a redirect.

### Results arriving after unmount

`ConnectBase` ran `init()` from a `useEffect` without cleanup, and both continuations of `connect()` ran whenever they eventually settled. Two ways that hurts: with the 5 s redirect windows above, a user who leaves the wallet in the meantime was thrown back to the wallet selection by the stale instance's `.catch`; and a wallet prompt approved after the user had moved on resolved the stale `getAccount`, whose success chain calls `doLogin` — `logout()`, `login()`, `setSession()`, `switchBlockchain()` — and could therefore log out a newer session and log the abandoned wallet in.

`init()` also treats a failing availability check as "not available" — `await isSupported()` sits in a `try`, and the `catch` leaves `supported` false — instead of leaving the spinner up forever with an unhandled rejection. The `try` also covers an implementation that throws synchronously, which the prop's `() => boolean | Promise<boolean>` type allows. No wallet hook rejects today, but the gate now depends on that call, so the failure mode is closed rather than documented.

`ConnectBase` now keeps a mounted ref and checks it at every point where a settled promise would act on the component or the app: it does not start `doLogin`, does not call `onLogin`, `onCancel` or `onSwitch`, and does not touch the sign hint once it is gone. The attempt is dropped rather than reported.

Two things deliberately still finish. A `doLogin` that already started runs to its end — it begins with `logout()`, so stopping in the middle would leave the user logged out and nothing else. And the wallet's own `signMessage` call is not cancelled; only the hint around it is suppressed, since the signature request lives in the wallet, not in this component.

### Found while touching `alby.hook.ts`

Two defects in code this pull request touches, both fixed here:

- `useAlby` memoized its return value with an empty dependency list, so `isEnabled` and the `sendPayment` closure stayed at the first render's `false`. The dependency list is now `[isEnabled]`.
- `waitForWebln()` checked ten times and then threw `Timeout` without looking once more after its last wait — the same off-by-one the two Tron hooks had. It matters more now, because `isAvailable()` wraps it and decides whether the install hint is shown. It now checks once more before giving up, which `enable()` benefits from as well.

### Behavior changes

- A browser without the wallet shows the install hint and nothing else: no connect error, no client error report. On desktop, TronLink and Trust (Tron) no longer open tronlink.org / trustwallet.com in a new tab, which their adapters did when the hidden connect ran without a wallet.
- For the two Tron wallets, that hint now appears up to 2 s after the screen opens, because the availability check waits that long for a late injected wallet.
- On a mobile browser, TronLink, Trust (Tron) and Phantom (iOS Safari) still open the page in the wallet app; if the app does not open, the user is back at the wallet selection after 5 s instead of seeing an error.
- Inside the Trust or TronLink app the connect still happens, and so it does for a wallet injected within the availability wait — about 100 ms for Alby, 2 s for the two Tron wallets. A TronLink extension that only appears after those 2 s is shown as unavailable, although the adapter's own 30 s wait would still have found it.
- Trust users whose `window.ethereum` belongs to another extension no longer see "Please install Trust!".

## Tests

- `src/__tests__/connect-base.test.tsx` (new): auto-connect gate, fallback switch, chain selection, abort / wallet switch / error handling, the login paths, three cases where the availability check itself fails (a rejected promise, with and without `autoConnect`, and an implementation that throws synchronously), and six unmount cases: no auto-connect, no fallback switch, no login, no `onCancel` — each of which fails without its guard — plus the two that document what deliberately continues (the wallet's signature request).
- `src/__tests__/connect-alby-redirect.test.tsx` (extended): the remaining `ConnectAlby` login branches.
- `src/__tests__/connect-tron-wallets.test.tsx` (new): `ConnectTrustTrx` and `ConnectTronLinkTrx` — install hint, mobile fallback, auto-connect, reconnect, connect error and signature forwarding.
- `src/hooks/wallets/__tests__/alby.hook.test.ts` (new): detection, the availability wait, enable / sign / payment including the `isEnabled` fix.
- `src/hooks/wallets/__tests__/solana-wallets.hook.test.ts` (new): Phantom and Trust (Solana) detection per adapter ready state, connect including the Phantom redirect, sign, transactions.
- `src/hooks/wallets/__tests__/tron-wallets.hook.test.ts` (new): Trust (Tron) and TronLink detection for injected wallet / in-app / mobile browser / desktop / no browser, the `isAvailable()` retries, connect including the app redirect, sign, transactions.

Coverage per touched file (`npm run test -- --coverage --collectCoverageFrom=<file>`):

| File | Statements | Branches | Functions | Lines |
| `src/components/home/connect-base.tsx` | 100 % | 100 % | 100 % | 100 % |
| `src/components/home/wallet/connect-alby.tsx` | 100 % | 100 % | 100 % | 100 % |
| `src/components/home/wallet/connect-tronlink-trx.tsx` | 100 % | 100 % | 100 % | 100 % |
| `src/components/home/wallet/connect-trust-trx.tsx` | 100 % | 100 % | 100 % | 100 % |
| `src/hooks/wallets/alby.hook.ts` | 100 % | 100 % | 100 % | 100 % |
| `src/hooks/wallets/phantom.hook.ts` | 100 % | 100 % | 100 % | 100 % |
| `src/hooks/wallets/tronlink-trx.hook.ts` | 100 % | 100 % | 100 % | 100 % |
| `src/hooks/wallets/trust-sol.hook.ts` | 100 % | 100 % | 100 % | 100 % |
| `src/hooks/wallets/trust-trx.hook.ts` | 100 % | 100 % | 100 % | 100 % |

Locally on Node 20: `npm run lint`, Prettier on all touched files, `npm run build:dev`, `npm run widget:dev`, and the full Jest suite (157 suites, 2439 tests) are green.

## Declared deviation: handbook coverage

Guideline: CONTRIBUTING.md › Handbook, "Every screen or flow a pull request changes has to be represented there" (a committed Playwright baseline per visual variant plus a `scripts/handbook/metadata.json` entry).

This PR changes the wallet connect flow: which of the existing states (install hint, connecting, connect error, back to the wallet selection) a user reaches depends now on the corrected wallet detection. It adds no new visual variant and does not change how any of these states look. None of these states has a baseline or a spec today; the existing login baselines (`login-process`, `user-flows`, `responsive`) authenticate with a session token and never render `ConnectBase`.

Reason for not adding baselines in this PR: every state this PR changes depends on the presence and timing of a browser wallet extension or a wallet in-app browser (injected, injected late, iOS Safari, inside the Trust or TronLink app), which the Playwright harness cannot reproduce; only the install hint without any extension would be capturable, and its appearance is unchanged. The state matrix is covered by the unit tests above instead. Whether this deviation is accepted is the reviewer's decision.

## Reported, not fixed here: the account after a wallet switch

When `getAccount` rejects with a `WalletSwitchError`, `connect()` switches the wallet and asks the new one for an account — but that call sits in the terminal `.catch()`, so its result goes nowhere: it never reaches `doLogin`, and a rejection has no handler. A BitBox user who picks the Ethereum entry on a Bitcoin-only device is switched to the Bitcoin wallet, selects an address, and is simply not logged in.

The code is unchanged by this pull request (identical on `develop`), but this pull request adds the first test that covers the path, and that test records the current behaviour. Repairing it means restructuring `connect()` so the retry re-enters the same login pipeline, which changes how every wallet switch behaves — a different change from the availability gate this pull request is about. Reported here for the reviewer to route, per CONTRIBUTING.md's rule on pre-existing defects.

## Reported, not fixed here: a login that has already started

Once `doLogin` has begun, it runs to completion even if `ConnectBase` unmounts in the meantime: `useWalletContext().login` fetches the sign message, asks the wallet to sign, calls `createSession`, sets the active wallet and blockchain, and reloads the user. A user who closes the connect screen while the wallet prompt is still open and then signs anyway therefore ends up logged in with that wallet, although the screen reported nothing.

This is not introduced by this pull request — before it, no continuation was guarded at all — and it is not this repository's wallet detection: stopping it properly means carrying a cancellation signal into `login()` in `src/contexts/wallet.context.tsx`, which every wallet shares, and deciding what a half-finished login should leave behind (`login()` starts after `logout()`, so an abort in the wrong place logs the user out and nothing else). Fixing that here would widen this pull request well beyond the connect gate it is about, so it is reported and left for the reviewer to route, per CONTRIBUTING.md's rule on pre-existing defects.

## Not part of this change

- Error classification for client error reports and anything outside this repository: the fix removes the source of the errors instead of reclassifying them.
- Full-stack E2E: the harness runs without a browser wallet extension, so the states this change turns on — injected, injected a moment late, inside a wallet's own app, redirected into one — cannot be reached there; the unit tests above cover them. The plain "no wallet at all" case would be reachable, but no full-stack spec opens an individual connect screen today (`e2e-stack/specs/auth.spec.ts` checks the wallet grid), and this change does not add the first one. No fakes were added or changed, so the reality declaration is unchanged.

</details>

### DIFF
--- a/src/__tests__/connect-alby-redirect.test.tsx
+++ b/src/__tests__/connect-alby-redirect.test.tsx
@@ -1,19 +1,22 @@
-// Component-level: ConnectAlby merges redirectPath + current search into the Alby
-// auth redirect's `redirect` query param via relativeUrl (personal-iban survival).
+// Component-level coverage for ConnectAlby: Alby auth redirect (personal-iban survival),
+// pubkey login, install-hint gate via isAvailable, Content error/loading branches, and
+// appParams / redirectPath guards.
 
 const mockEnable = jest.fn();
 const mockSignMessage = jest.fn();
-const mockIsInstalled = jest.fn();
+const mockIsAvailable = jest.fn();
 const mockRedirectPath = jest.fn();
 const mockAppParams = jest.fn();
 const mockOnCancel = jest.fn();
 const mockOnLogin = jest.fn();
 const mockOnSwitch = jest.fn();
+const mockLogin = jest.fn();
 
 jest.mock('@dfx.swiss/react', () => ({
   Blockchain: { LIGHTNING: 'Lightning' },
   useAuthContext: () => ({ session: undefined }),
   useSessionContext: () => ({ logout: jest.fn() }),
+  useUserContext: () => ({ user: undefined }),
 }));
 
 jest.mock('@dfx.swiss/react-components', () => ({
@@ -24,8 +27,18 @@ jest.mock('@dfx.swiss/react-components', () => ({
     </button>
   ),
   StyledButtonColor: { GRAY_OUTLINE: 'gray-outline' },
-  StyledButtonWidth: { MIN: 'min' },
+  StyledButtonWidth: { MIN: 'min', SM: 'sm' },
   StyledLoadingSpinner: () => null,
+  StyledVerticalStack: ({ children }: any) => <div>{children}</div>,
+  StyledLink: ({ label }: any) => <a>{label}</a>,
+}));
+
+jest.mock('react-i18next', () => ({
+  Trans: ({ children }: any) => <>{children}</>,
+}));
+
+jest.mock('../hooks/report-displayed-error.hook', () => ({
+  useReportDisplayedError: () => undefined,
 }));
 
 jest.mock('../config/api', () => ({
@@ -50,7 +63,7 @@ jest.mock('../contexts/wallet.context', () => ({
   WalletBlockchains: { Alby: ['Lightning'] },
   supportsBlockchain: () => true,
   useWalletContext: () => ({
-    login: jest.fn(),
+    login: mockLogin,
     setSession: jest.fn(),
     switchBlockchain: jest.fn(),
     activeWallet: undefined,
@@ -59,7 +72,7 @@ jest.mock('../contexts/wallet.context', () => ({
 
 jest.mock('../hooks/wallets/alby.hook', () => ({
   useAlby: () => ({
-    isInstalled: mockIsInstalled,
+    isAvailable: mockIsAvailable,
     enable: mockEnable,
     signMessage: mockSignMessage,
   }),
@@ -73,7 +86,7 @@ jest.mock('../util/utils', () => {
   };
 });
 
-import { act, render, waitFor } from '@testing-library/react';
+import { act, render, screen, waitFor } from '@testing-library/react';
 import { createRef } from 'react';
 import ConnectAlby from '../components/home/wallet/connect-alby';
 import { WalletType } from '../contexts/wallet.context';
@@ -85,10 +98,11 @@ describe('ConnectAlby login redirect', () => {
   beforeEach(() => {
     jest.clearAllMocks();
     capturedLocation = undefined;
-    mockIsInstalled.mockResolvedValue(true);
+    mockIsAvailable.mockResolvedValue(true);
     mockEnable.mockResolvedValue({ node: { alias: 'getalby.com' } });
     mockAppParams.mockReturnValue({});
     mockOnCancel.mockClear();
+    mockLogin.mockResolvedValue(undefined);
 
     locationStub = {
       href: 'http://localhost/connect',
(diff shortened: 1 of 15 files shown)


TaprootFreakAI (approved):
EN:
Without a wallet in the browser, auto-connect no longer runs, so the false connect errors behind the install hint go away; real connection failures stay ERROR.

DE:
Ohne Wallet im Browser wird nicht mehr automatisch verbunden, deshalb verschwinden die falschen Connect-Fehler hinter dem Install-Hinweis; echte Verbindungsfehler bleiben ERROR.

TaprootFreak (approved):

